PR #17, opened 2021-04-28
Cross-compilation and separation of codecs from serializer

MarconZet:


project/Dependencies.scala
```diff
@@ -18,6 +18,8 @@ object Dependencies {
 
   val reflections = "net.oneandone.reflections8" % "reflections8" % "0.11.7"
 
+  val logger = "org.slf4j" % "slf4j-simple" % "1.7.30"
+
```
PawelLipski, line 22: Pls bump Borer to 1.7.1 for 2.13 only (if not easily doable then again, drop 2.12 support as for now)

build.sbt
```diff
@@ -1,31 +1,52 @@
...
+    crossScalaVersions := Nil,
+    publish / skip := true
   )
 
+lazy val core = (projectMatrix in file("core"))
+  .dependsOn(codex)
+  .settings(commonSettings)
+  .jvmPlatform(scalaVersions = supportedScalaVersions)
+
+lazy val codex = (projectMatrix in file("codex"))
```
PawelLipski, line 48: It's `codecs` aka `COders-DECoderS` ;) codex is well, https://en.wikipedia.org/wiki/Codex

project/Dependencies.scala
PawelLipski: Opened https://github.com/sirthias/borer/issues/389, let's see what happens

.gitignore
```diff
@@ -3,6 +3,8 @@
 
 .bsp/
 .idea/
+.sbt/
+null/
```
PawelLipski, line 7: Hmm where's that directory coming from? 🤔

project/plugins.sbt
```diff
@@ -1 +1,2 @@
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.2")
+addSbtPlugin("com.eed3si9n" % "sbt-projectmatrix" % "0.7.0")
```
PawelLipski, line 2: Nice! didn't know that, I used to define subproject for each Scala version manually

build.sbt
```diff
@@ -1,31 +1,52 @@
...
-      "-Ywarn-dead-code",
-      "-Ywarn-unused:-imports,_",
-      "-unchecked"),
-    libraryDependencies ++= deps
+    crossScalaVersions := Nil,
+    publish / skip := true
   )
 
+lazy val core = (projectMatrix in file("core"))
+  .dependsOn(codex)
```
PawelLipski, line 44: I guess the name `core` isn't adequate (can be misleading, I'd intuitively say that it's rather non-core that should depend on core, not the other way round)... maybe let's use the near-target names `borer-akka-serializer` and `borer-extra-codecs` then for the modules

build.sbt
```diff
@@ -1,31 +1,52 @@
...
+    "-Ybackend-parallelism",
+    "8",
+    "-Ywarn-dead-code",
+    "-Ywarn-unused:-imports,_",
+    "-unchecked"),
+  libraryDependencies ++= deps
+)
+
+
 lazy val root = (project in file("."))
```
PawelLipski, line 35: Nit: does project-matrix allow of getting rid of this top-level boilerplate module somehow?

.gitignore
MarconZet: Judging by internal folder structure, it was created by `sbt-projectmatrix` plugin

build.sbt
MarconZet: According to the example on the plugins Github page it does. I decided to preserve it in case it turns out to be needed again.

MarconZet:
After the project restructure the easiest way to get this library into another project is by using a local repository.
Publish locally using `sbt publishLocal` and include library dependencies:
```
libraryDependencies += "org.virtuslab" %% "borer-akka-serializer" % "0.1.0-SNAPSHOT"
libraryDependencies += "org.virtuslab" %% "borer-extra-codecs" % "0.1.0-SNAPSHOT"
```

build.sbt
MarconZet: Now that I think about it, the serializer doesn't really depend on codecs, test for it do.

build.sbt
```diff
@@ -25,7 +28,34 @@ lazy val root = (project in file("."))
...
-    libraryDependencies ++= deps
-  )
+  libraryDependencies ++= standardDeps)
+
+//lazy val root = (project in file("."))
+//  .aggregate(serializer.projectRefs ++ codecs.projectRefs: _*)
+//  .settings(name := "akka-safer-serializer", crossScalaVersions := Nil, publish / skip := true)
+
+lazy val serializer = (projectMatrix in file("serializer"))
+  .dependsOn(codecs)
```
PawelLipski, line 38: Ok so pls leave the dependency only in Test scope if possible...

build.sbt
```diff
@@ -25,7 +28,34 @@ lazy val root = (project in file("."))
       "-Ywarn-dead-code",
       "-Ywarn-unused:-imports,_",
       "-unchecked"),
-    libraryDependencies ++= deps
-  )
+  libraryDependencies ++= standardDeps)
+
+//lazy val root = (project in file("."))
```
PawelLipski, line 33: Remove?

project/Dependencies.scala
```diff
@@ -1,33 +1,36 @@
...
-
-  val borerCore = "io.bullet" %% "borer-core" % borerVersion
-  val borerDerivation = "io.bullet" %% "borer-derivation" % borerVersion
-  val borerAkka = "io.bullet" %% "borer-compat-akka" % borerVersion
-
   val reflections = "net.oneandone.reflections8" % "reflections8" % "0.11.7"
+  val logger = "org.slf4j" % "slf4j-simple" % "1.7.30"
 
-  val deps = Seq(
+  val standardDeps = Seq(
```
PawelLipski, line 17: Nit: I'd say `commonDeps` or `sharedDeps`

serializer/src/main/scala-2.13/org/virtuslab/akkasaferserializer/RuntimeReflections.scala
```diff
@@ -0,0 +1,15 @@
+package org.virtuslab.akkasaferserializer
+
+import org.reflections8.Reflections
+
+import scala.jdk.CollectionConverters._
```
PawelLipski, line 5: Nit: there is an import that's portable b/w both Scala 2.12 and 2.13 (although deprecated in the latter)... `import scala.collection.JavaConverters._`?

PawelLipski: So duplicating code b/w scala-2.12 and scala-2.13 can be avoided

project/plugins.sbt
PawelLipski: ...although probably not needed now that you've used the `VirtualAxes` trick (another thing I didn't know, looks pretty useful)

serializer/src/test/scala-2.13/org/virtuslab/akkasaferserializer/CrossBuildSpec.scala
```diff
@@ -0,0 +1,25 @@
+package org.virtuslab.akkasaferserializer
+
+import io.bullet.borer.derivation.MapBasedCodecs.deriveCodec
+import io.bullet.borer.{Cbor, Codec}
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class CrossBuildSpec extends AnyFlatSpec with Matchers {
```
PawelLipski, line 8: <3

build.sbt
```diff
@@ -25,7 +28,34 @@ lazy val root = (project in file("."))
...
+
+lazy val serializer = (projectMatrix in file("serializer"))
+  .dependsOn(codecs)
+  .settings(name := "borer-akka-serializer")
+  .settings(commonSettings)
+  .settings(libraryDependencies ++= {
+    virtualAxes.value.collectFirst({ case x: ScalaVersionAxis => x.value }).fold(Seq[ModuleID]()) {
+      case "2.13" => scala213Deps
+      case "2.12" => scala212Deps
+    }
```
PawelLipski, line 45: Nit:
```suggestion
    virtualAxes.value.collectFirst { case x: ScalaVersionAxis => x.value }.map {
      case "2.13" => scala213Deps
      case "2.12" => scala212Deps
    }.getOrElse(Seq.empty)```

Comment (PawelLipski) [reply build.sbt:45]: I find `fold` on Options hard to read, `map` + `getOrElse` is easier to reason about

File: build.sbt (hunk 6 above)
Comment (PawelLipski) [reply build.sbt]: Oh ok let's leave then

**Review by PawelLipski — APPROVED** [review]

File: serializer/src/main/scala-2.13/org/virtuslab/akkasaferserializer/RuntimeReflections.scala (hunk 10 above)
Comment (MarconZet) [reply serializer/src/main/scala-2.13/org/virtuslab/akkasaferserializer/RuntimeReflections.scala:5]:
```
[error] [...]\RuntimeReflections.scala:14:86: object ImplicitConversions in package convert is deprecated (since 2.13.0): Use `scala.jdk.CollectionConverters` instead
[error]   def findAllObjects[T](cl: Class[T]): Seq[Class[_ <: T]] = reflections.getSubTypesOf(cl).toSeq
[error]                                                                                      ^
[error] one error found
[error] (serializer / Compile / compileIncremental) Compilation failed
```
Deprecated but doesn't compile

MarconZet: Ok, found it.
It's becaouse of flag `-Xfatal-warnings`.
Should we remove it?

PawelLipski: Ouch, right... remove `-deprecation` maybe, if it doesn't help, then leave as is

MarconZet: No luck. It seems that's the way it's supposed to work.

PawelLipski:
Pls bump borer deps for both Scala versions to 1.7.2 once they make their way to Maven Central

PR #19, opened 2021-04-30
Compiler plugin for checking whether the specified serializer is used

MarconZet:


build.sbt
```diff
@@ -59,4 +59,8 @@ lazy val codecs = (projectMatrix in file("codecs"))
   })
   .jvmPlatform(scalaVersions = supportedScalaVersions)
 
-// See https://www.scala-sbt.org/1.x/docs/Using-Sonatype.html for instructions on how to publish to Sonatype.
+lazy val plugin = (projectMatrix in file("plugin"))
+  .settings(name := "safer-serializer-plugin")
+  .settings(commonSettings)
+  .settings(libraryDependencies ++= Seq(scalaCompiler213, scalaLibrary213, scalaReflect213))
```
PawelLipski, line 65: Ok... so as for now, the plugin is 2.13 only right?

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SerializerTrait.scala
```diff
@@ -0,0 +1,3 @@
+package org.virtuslab.akkasaferserializer
+
+class SerializerTrait extends scala.annotation.ConstantAnnotation
```
PawelLipski, line 3: I'm testing that now... there's a small problem here, this trait should be both on the classpath of the plugin AND of the project itself ;) so just adding `compilerPlugin("org.virtuslab" %% "safer-serializer-plugin" % "0.1.0-SNAPSHOT")` is not enough, I ended up with `scala.ScalaReflectionException: object org.virtuslab.akkasaferserializer.SerializerTrait in compiler mirror not found.`

PawelLipski: Probably `SerializerTrait` should then be extracted to a separate project... or maybe, the FQCN of the trait can be passed via `-P...` option to Scala compiler instead?

build.sbt
PawelLipski: 2.12 would be useful (even instead of 2.13, if cross-compilation is non-trivial) since as of today, Hydra is still running on 2.12

plugin/src/test/resources/SingleBehaviorYes.scala
```diff
@@ -0,0 +1,11 @@
+package org.random.project
+
+import akka.actor.typed.Behavior
+import org.virtuslab.akkasaferserializer.SerializerTrait
+
+object SingleBehaviorYes {
+  @SerializerTrait
```
PawelLipski, line 7: Hmmm actually for some reason, I'm getting `does not extend annotated trait` for all classes that are type params of Behavior, even after annotating `CborSerializable` as `@SerializerTrait` :/

Maybe it's the problem that I needed to add the plugin as both regular dependency (`"org.virtuslab" %% "safer-serializer-plugin" % "0.1.0-SNAPSHOT"`) and compiler plugin (`compilerPlugin("org.virtuslab" %% "safer-serializer-plugin" % "0.1.0-SNAPSHOT")`)? Then maybe `SerializerTrait` referred to from `StdPhase` is not recognized as the same thing as `SerializerTrait` present on the project classpath?...

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,47 @@
...
+
+        val roots = body.collect {
+          case x: ClassDef if x.symbol.annotations.exists(_.atp =:= typeOf[SerializerTrait]) => x.symbol.tpe
+        }.toSet
+
+        val typesToCheck = body.collect {
+          case x: TypeTree if compareGenerics(x.tpe, typeOf[Behavior[Nothing]]) => x.tpe.typeArgs
+        }.flatten
+
+        @tailrec
```
PawelLipski, line 27: Hmmm recursive BFS? 🤔

PawelLipski: Oh okay, I see, makes sense

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,47 @@
...
+        def typesBfs(prev: List[Type], targets: Set[Type]): Boolean = {
```
PawelLipski, line 28: Pls use some more descriptive name... what's the meaning of the returned value?

plugin/src/test/resources/SingleBehaviorYes.scala
MarconZet: I just did that in my test project and it worked. Maybe the problem was with faulty BFS. I have already replaced it with something more efficient. Could you check again with newest version?
```
addCompilerPlugin("org.virtuslab" %% "safer-serializer-plugin" % "0.1.0-SNAPSHOT")
libraryDependencies += "org.virtuslab" %% "safer-serializer-plugin" % "0.1.0-SNAPSHOT"
```

MarconZet: Ok, I take that back. It doesn't work when there are multiple source files.

MarconZet: > Ok, I take that back. It doesn't work when there are multiple source files.

Fixed

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SerializerTrait.scala
MarconZet: I think annotation with double import is the cleanest option. I believe annotation is easier to use than `-P` scalac option and separating plugin form annotation neither reduce boilerplate code nor improve functionality.

PawelLipski: Okay! lemme try now then...

plugin/src/test/resources/SingleBehaviorYes.scala
PawelLipski: Hmm I've now added a plugin built locally from `fa2c5be80bd81b6880d71fdef831cad156a85aea` to the project (Scala 2.12):

```
    "org.virtuslab" %% "safer-serializer-plugin" % "0.1.0-SNAPSHOT",
    compilerPlugin("org.virtuslab" %% "safer-serializer-plugin" % "0.1.0-SNAPSHOT")
```

I'm getting:


```
[error] ## Exception when compiling 297 sources to /home/plipski/.../target/scala-2.12/classes
[error] java.lang.NoSuchMethodError: 'scala.tools.nsc.reporters.FilteringReporter scala.tools.nsc.Global.reporter()'
[error] org.virtuslab.akkasaferserializer.CheckTypesPluginComponent$$anon$1.apply(CheckTypesPluginComponent.scala:22)
[error] scala.tools.nsc.Global$GlobalPhase.applyPhase(Global.scala:454)
[error] scala.tools.nsc.Global$GlobalPhase.run(Global.scala:402)
[error] scala.tools.nsc.Global$Run.compileUnitsInternal(Global.scala:1511)
[error] scala.tools.nsc.Global$Run.compileUnits(Global.scala:1495)
[error] scala.tools.nsc.Global$Run.compileSources(Global.scala:1488)
[error] scala.tools.nsc.Global$Run.compileFiles(Global.scala:1596)
[error] xsbt.CachedCompiler0.run(CompilerBridge.scala:163)
[error] xsbt.CachedCompiler0.run(CompilerBridge.scala:134)
[error] xsbt.CompilerBridge.run(CompilerBridge.scala:39)
[error] sbt.internal.inc.AnalyzingCompiler.compile(AnalyzingCompiler.scala:92)
[error] sbt.internal.inc.MixedAnalyzingCompiler.$anonfun$compile$7(MixedAnalyzingCompiler.scala:186)
[error] scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
[error] sbt.internal.inc.MixedAnalyzingCompiler.timed(MixedAnalyzingCompiler.scala:241)
[error] sbt.internal.inc.MixedAnalyzingCompiler.$anonfun$compile$4(MixedAnalyzingCompiler.scala:176)
[error] sbt.internal.inc.MixedAnalyzingCompiler.$anonfun$compile$4$adapted(MixedAnalyzingCompiler.scala:157)
[error] sbt.internal.inc.JarUtils$.withPreviousJar(JarUtils.scala:239)
[error] sbt.internal.inc.MixedAnalyzingCompiler.compileScala$1(MixedAnalyzingCompiler.scala:157)
[error] sbt.internal.inc.MixedAnalyzingCompiler.compile(MixedAnalyzingCompiler.scala:204)
[error] sbt.internal.inc.IncrementalCompilerImpl.$anonfun$compileInternal$1(IncrementalCompilerImpl.scala:573)
[error] sbt.internal.inc.IncrementalCompilerImpl.$anonfun$compileInternal$1$adapted(IncrementalCompilerImpl.scala:573)
[error] sbt.internal.inc.Incremental$.$anonfun$apply$5(Incremental.scala:173)
[error] sbt.internal.inc.Incremental$.$anonfun$apply$5$adapted(Incremental.scala:171)
[error] sbt.internal.inc.Incremental$$anon$2.run(Incremental.scala:458)
[error] sbt.internal.inc.IncrementalCommon$CycleState.next(IncrementalCommon.scala:116)
[error] sbt.internal.inc.IncrementalCommon$$anon$1.next(IncrementalCommon.scala:56)
[error] sbt.internal.inc.IncrementalCommon$$anon$1.next(IncrementalCommon.scala:52)
[error] sbt.internal.inc.IncrementalCommon.cycle(IncrementalCommon.scala:261)
[error] sbt.internal.inc.Incremental$.$anonfun$incrementalCompile$8(Incremental.scala:413)
[error] sbt.internal.inc.Incremental$.withClassfileManager(Incremental.scala:498)
[error] sbt.internal.inc.Incremental$.incrementalCompile(Incremental.scala:400)
[error] sbt.internal.inc.Incremental$.apply(Incremental.scala:165)
[error] sbt.internal.inc.IncrementalCompilerImpl.compileInternal(IncrementalCompilerImpl.scala:573)
[error] sbt.internal.inc.IncrementalCompilerImpl.$anonfun$compileIncrementally$1(IncrementalCompilerImpl.scala:491)
[error] sbt.internal.inc.IncrementalCompilerImpl.handleCompilationError(IncrementalCompilerImpl.scala:332)
[error] sbt.internal.inc.IncrementalCompilerImpl.compileIncrementally(IncrementalCompilerImpl.scala:420)
[error] sbt.internal.inc.IncrementalCompilerImpl.compile(IncrementalCompilerImpl.scala:137)
[error] sbt.Defaults$.compileIncrementalTaskImpl(Defaults.scala:2176)
[error] sbt.Defaults$.$anonfun$compileIncrementalTask$2(Defaults.scala:2133)
[error] sbt.internal.io.Retry$.apply(Retry.scala:40)
[error] sbt.internal.io.Retry$.apply(Retry.scala:23)
[error] sbt.internal.server.BspCompileTask$.compute(BspCompileTask.scala:31)
[error] sbt.Defaults$.$anonfun$compileIncrementalTask$1(Defaults.scala:2129)
[error] scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error] sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
[error] sbt.std.Transform$$anon$4.work(Transform.scala:68)
[error] sbt.Execute.$anonfun$submit$2(Execute.scala:282)
[error] sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:23)
[error] sbt.Execute.work(Execute.scala:291)
[error] sbt.Execute.$anonfun$submit$1(Execute.scala:282)
[error] sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:265)
[error] sbt.CompletionService$$anon$2.call(CompletionService.scala:64)
[error] java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[error] java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
[error] java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[error] java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[error] java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[error] java.base/java.lang.Thread.run(Thread.java:834)
[error]            
[error] stack trace is suppressed; run last ... / Compile / compileIncremental for the full output
[error] (.... / Compile / compileIncremental) java.lang.NoSuchMethodError: 'scala.tools.nsc.reporters.FilteringReporter scala.tools.nsc.Global.reporter()'
[error] Total time: 115 s (01:55), completed 6 May 2021, 12:32:43
```

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SerializerTrait.scala
PawelLipski: Under Scala 2.13, I'm getting a false positive error:

```
[error] /home/plipski/hydra-backend/services/src/main/scala/hydra/wallets/WalletActorProtocol.scala:21:18: hydra.wallets.WalletActorProtocol.commands.WalletCommand does not extend annotated trait
[error]     sealed trait WalletCommand extends CborSerializable {
[error]                  ^
```

PawelLipski: Where `CborSerializable` is defined as:

```
package hydra

import org.virtuslab.akkasaferserializer.SerializerTrait

@SerializerTrait
trait CborSerializable
```

PawelLipski: And WalletCommand is:

```scala
private object WalletActorProtocol {

  object commands {

    sealed trait WalletCommand extends CborSerializable {
      def walletId: WalletId
      val replyTo: ActorRef[WalletResponse]
    }
// ...
```

plugin/src/main/scala/org/virtuslab/akkasaferserializer/CheckTypesPluginComponent.scala
```diff
@@ -0,0 +1,29 @@
...
+    new StdPhase(prev) {
+      private var runOnce = true
+
+      override def apply(unit: global.CompilationUnit): Unit = {
+        if (runOnce) {
+          val roots = gatherTypes.roots
+          val leafs = gatherTypes.leafs
+
+          leafs.find(x => !roots.exists(y => x <:< y)) match {
+            case Some(tp) => reporter.error(tp.typeSymbol.pos, s"${tp.toString()} does not extend annotated trait")
```
PawelLipski, line 22: Pls make this error message clearer (what the trait should be annotated with + what will happen because of that missing trait and why)

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SerializerTrait.scala
PawelLipski: Not a matter of `private object`, with just `object` it's still the same false-positive error... maybe a matter of nested objects?

plugin/src/test/resources/SingleBehaviorYes.scala
MarconZet: This is very bad, can't reproduce. May I ask sbt version?

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SerializerTrait.scala
MarconZet: I think the plugin handles partial compilation very badly. Try `sbt clean` before `sbt compile`

plugin/src/test/resources/SingleBehaviorYes.scala
PawelLipski: sbt is `1.4.4`. I can later provide additional details if needed

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SerializerTrait.scala
PawelLipski: Hmmm no change after a clean... maybe you could add some additional log messages to the plugin? So that I can provide more details about what's important in the context

plugin/src/test/resources/SingleBehaviorYes.scala
MarconZet: > Hmm I've now added a plugin built locally from `fa2c5be80bd81b6880d71fdef831cad156a85aea` to the project 

@PawelLipski Fixed by fc5430c5ccf42b965acc2383b2d5f0e03d51715e

project/Dependencies.scala
```diff
@@ -1,6 +1,9 @@
 import sbt._
 
 object Dependencies {
+  lazy val scala212 = "2.12.13"
```
PawelLipski, line 4: Nit: doesn't need to be lazy

PawelLipski: (it'd need if the initialization order was fishy and you risked the value being referenced before initialization, but here it's rather clear AFAICS)

plugin/src/test/scala/org/virtuslab/akkasaferserializer/compiler/TestCompilerSpec.scala
```diff
@@ -0,0 +1,18 @@
+package org.virtuslab.akkasaferserializer.compiler
+
+import org.scalatest.flatspec.AnyFlatSpecLike
+import org.scalatest.matchers.should.Matchers
+
+class TestCompilerSpec extends AnyFlatSpecLike with Matchers {
```
PawelLipski, line 6: Wow testing the test code 😅

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SerializerTrait.scala
```diff
@@ -0,0 +1,3 @@
+package org.virtuslab.akkasaferserializer
+
+class SerializerTrait extends scala.annotation.StaticAnnotation
```
PawelLipski, line 3: Actually... the name `SerializerTrait` is misleading a bit, since `Serializer` suggests a class like `akka.serialization.Serializer` that actively does serialize some 3rd party object... while this trait is really used to mark **serializable objects**.

PawelLipski: So... maybe `Serializability(Marker?)Trait`?

plugin/src/test/resources/MySer.scala
```diff
@@ -0,0 +1,6 @@
+package org.random.project
+
+import org.virtuslab.akkasaferserializer.SerializerTrait
+
+@SerializerTrait
+trait MySer
```
PawelLipski, line 6: Nit: pls expand `Ser` 🧀

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  override def newPhase(prev: Phase): Phase =
+    new StdPhase(prev) {
+      override def apply(unit: global.CompilationUnit): Unit = {
+        val body = unit.body
+        val reporter = CrossVersionReporter(global)
+
+        rootsCache = body
+          .collect {
+            case x: TypeTree if compareGenerics(x.tpe, typeOf[Behavior[Nothing]]) => x.tpe.typeArgs
```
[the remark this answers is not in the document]
PawelLipski: There's still a false positive error, but this time it's expected: `akka.projection.ProjectionBehavior.Command` and `akka.NotUsed` are false flagged. Pls whitelist some common Akka classes like that (I can also think of e.g. `akka.Done`).

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,61 @@
...
+            } else {
+              superclassDfs(next) match {
+                case Some(tp) =>
+                  if (pluginOptions.verbose)
+                    reporter.echo(s"SSP: Found new root: ${tp.typeSymbol.fullName}")
+                  tp :: roots
+                case None =>
+                  reporter.error(
+                    next.typeSymbol.pos,
+                    s"""${next.toString()} is used as Akka message but does is not annotated or extends annotated trait
```
PawelLipski, line 38: `does is` whoops

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,61 @@
...
+                     |Annotate it or its superclass with @${classOf[SerializerTrait].getName}
+                     |This may cause an unexpected use of java serialization during runtime
```
PawelLipski, line 40:
```suggestion
                     |Passing an object NOT extending ${...} as a message may cause Akka to fall back to Java serialization during runtime.
```

PawelLipski, line 40:
```suggestion
                     |Passing an object NOT extending ${...} as a message may cause Akka to fall back to Java serialization during runtime.
```

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
PawelLipski:
```suggestion
                    s"""${next.toString()} is used as Akka message but does not extend a trait annotated with ${classOf[SerializerTrait].getName}
```

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,61 @@
...
+                     |""".stripMargin)
+                  rootsCache
+              }
+            }
+
+          }
+
+      }
+      private def compareGenerics(t1: Type, t2: Type): Boolean = {
```
PawelLipski, line 49: I'd rather say `areGenericsEquivalent`

PawelLipski: Coz `compare...` doesn't say what the result is supposed to mean (or it might suggest that the result is an int, see `java.lang.Comparable#compareTo`

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,61 @@
...
+          }
+          .flatten
+          .foldRight(rootsCache) { (next, roots) =>
+            if (roots.exists(next <:< _)) {
+              roots
+            } else {
+              superclassDfs(next) match {
+                case Some(tp) =>
+                  if (pluginOptions.verbose)
+                    reporter.echo(s"SSP: Found new root: ${tp.typeSymbol.fullName}")
```
PawelLipski, line 33: Use `${classOf[SaferSerializerPluginComponent].getSimpleName}` instead of `SSP`, that's much more refactor-friendly

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import scala.tools.nsc.plugins.PluginComponent
+import scala.tools.nsc.{Global, Phase}
+
+class SaferSerializerPluginComponent(val pluginOptions: PluginOptions, val global: Global) extends PluginComponent {
+  import global._
+  override val phaseName: String = "akka-safer-serializer-gather"
+  override val runsAfter: List[String] = List("refchecks")
+
+  var rootsCache: List[Type] = List()
```
PawelLipski, line 13: Actually... what's the lifetime of a single `SaferSerializerPluginComponent` object? Is it e.g. cached b/w separate invocations of `compile` in a single `sbt` session? Maybe it'd be a better idea to make the phase completely stateless (everything kept in local vars & method params and not fields?)

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,61 @@
...
+          .collect {
+            case x: TypeTree if compareGenerics(x.tpe, typeOf[Behavior[Nothing]]) => x.tpe.typeArgs
+          }
+          .flatten
+          .foldRight(rootsCache) { (next, roots) =>
+            if (roots.exists(next <:< _)) {
+              roots
+            } else {
+              superclassDfs(next) match {
+                case Some(tp) =>
```
PawelLipski, line 31: Nit: pls expand `tp` to sth more meaningful (`superclassType`?)

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,61 @@
...
+              }
+            }
+
+          }
+
+      }
+      private def compareGenerics(t1: Type, t2: Type): Boolean = {
+        t1.prefix =:= t2.prefix && t1.typeSymbol == t2.typeSymbol
+      }
+      private def superclassDfs(tp: Type): Option[Type] = {
```
PawelLipski, line 52: `findSuperclassAnnotatedWithSerializabilityTrait`?

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,61 @@
...
+      override def apply(unit: global.CompilationUnit): Unit = {
+        val body = unit.body
+        val reporter = CrossVersionReporter(global)
+
+        rootsCache = body
+          .collect {
+            case x: TypeTree if compareGenerics(x.tpe, typeOf[Behavior[Nothing]]) => x.tpe.typeArgs
+          }
+          .flatten
+          .foldRight(rootsCache) { (next, roots) =>
```
PawelLipski, line 26: Hmm is `roots` actually the best name here? IIUC, maybe it could be sth like `annotatedTraits`?

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPlugin.scala
```diff
@@ -0,0 +1,22 @@
...
+  private val gatherTypes = new SaferSerializerPluginComponent(pluginOptions, global)
+  override val components: List[PluginComponent] = List(gatherTypes)
+
+  override def init(options: List[String], error: String => Unit): Boolean = {
+    pluginOptions.verbose = options.exists(_.equals("verbose"))
+    true
+  }
+
+  override val optionsHelp: Option[String] = Some("""
+      |verbose - print additional info about leafs and roots
```
PawelLipski, line 20: Hmmm it's pretty unclear what `leafs and roots` correspond to here

PawelLipski: I'd either skip it completely (just `print additional diagnostic info`) or provide some clearer explanation (`print additional info about detected serializable classes`?)

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPlugin.scala
```diff
@@ -0,0 +1,22 @@
...
+class SaferSerializerPlugin(override val global: Global) extends Plugin {
+  override val name: String = "safer-serializer-plugin"
+  override val description: String = "checks for Behavior[A] and ActorRef[A]"
+  private val pluginOptions = new PluginOptions(false)
+
+  private val gatherTypes = new SaferSerializerPluginComponent(pluginOptions, global)
+  override val components: List[PluginComponent] = List(gatherTypes)
+
+  override def init(options: List[String], error: String => Unit): Boolean = {
+    pluginOptions.verbose = options.exists(_.equals("verbose"))
```
PawelLipski, line 15: Nit: `equals` is uncommon in Scala code, just `_ == "verbose"`

PawelLipski: (likely not a problem here, but `equals` doesn't check for call target being null, while `==` does)

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPlugin.scala
```diff
@@ -0,0 +1,22 @@
+package org.virtuslab.akkasaferserializer
+
+import scala.tools.nsc.Global
+import scala.tools.nsc.plugins.{Plugin, PluginComponent}
+
+class SaferSerializerPlugin(override val global: Global) extends Plugin {
+  override val name: String = "safer-serializer-plugin"
+  override val description: String = "checks for Behavior[A] and ActorRef[A]"
+  private val pluginOptions = new PluginOptions(false)
```
PawelLipski, line 9: Nit: name the `false` param

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPlugin.scala
```diff
@@ -0,0 +1,22 @@
+package org.virtuslab.akkasaferserializer
+
+import scala.tools.nsc.Global
+import scala.tools.nsc.plugins.{Plugin, PluginComponent}
+
+class SaferSerializerPlugin(override val global: Global) extends Plugin {
+  override val name: String = "safer-serializer-plugin"
+  override val description: String = "checks for Behavior[A] and ActorRef[A]"
```
PawelLipski, line 8: I'd make it the description more generic, since there'll be surely many more classes to consider: maybe sth like `checks that the specified Akka serialization is applied to all messages, event and persistent state classes`

build.sbt
```diff
@@ -43,4 +42,16 @@ lazy val codecs = (projectMatrix in file("codecs"))
   .settings(libraryDependencies ++= borerDeps)
   .jvmPlatform(scalaVersions = supportedScalaVersions)
 
-// See https://www.scala-sbt.org/1.x/docs/Using-Sonatype.html for instructions on how to publish to Sonatype.
+lazy val plugin = (projectMatrix in file("plugin"))
+  .settings(name := "safer-serializer-plugin")
+  .settings(commonSettings)
```
PawelLipski, line 47: Some `commonDeps` like `reflections` aren't needed here

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
MarconZet: From my experiments, it's a single compilation run. That's why it's called cache, it's completely optional. We could remove it without any repercussion, besides a longer compilation time. In most projects, there will only be one trait, and most messages will already be extending it, so performance gain may be significant.

PawelLipski: Ok, if so then let's leave as is

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,64 @@
...
+                  if (pluginOptions.verbose)
+                    reporter.echo(s"SSP: Found new root: ${tp.typeSymbol.fullName}")
+                  tp :: roots
+                case None =>
+                  reporter.error(
+                    next.typeSymbol.pos,
+                    s"""${next.toString()} is used as Akka message but does is not annotated or extends annotated trait
+                     |Annotate it or its superclass with @${classOf[SerializerTrait].getName}
+                     |This may cause an unexpected use of java serialization during runtime
+                     |""".stripMargin)
```
PawelLipski, line 47: Also, the error messages aren't separated from each other:

image

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  override def newPhase(prev: Phase): Phase =
+    new StdPhase(prev) {
+      override def apply(unit: global.CompilationUnit): Unit = {
+        val body = unit.body
+        val reporter = CrossVersionReporter(global)
+
+        val generics = Seq(
+          typeOf[akka.actor.typed.Behavior[_]],
+          typeOf[akka.persistence.typed.scaladsl.ReplyEffect[Any, _]],
```
PawelLipski, line 23: Would it be possible to make `akka-persistence` (same for `akka-projection`) hmmm an _optional dependency_? As in, we'd check `ReplyEffect` type params if `akka-persistence` is already on the classpath, but otherwise ignore it? 

Note that making it a _provided dependency_ is not the same: now if `akka-persistence` is **not** on the classpath, we'd likely get some ClassNotFound error.

Maybe relying on reflection would be a better idea here actually? so that you find `akka.persistence.typed.scaladsl.ReplyEffect` class by name?

build.sbt
MarconZet: Dependencies management is one huge mess right now. I believe disentangling them qualifies for a different PR.

PawelLipski: Okay pls open an issue then with `build` label

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,64 @@
...
+                    reporter.echo(
+                      s"${classOf[SaferSerializerPlugin].getSimpleName}: Found new annotated trait: ${annotatedType.typeSymbol.fullName}")
+                  annotatedType :: annotatedTraits
+                case None =>
+                  reporter.error(
+                    next.typeSymbol.pos,
+                    s"""${next.toString()} is used as Akka message but does not extend a trait annotated with ${classOf[
+                      SerializabilityTrait].getName}.
+                     |Annotate this class or one of the traits/classes it extends with @${classOf[SerializabilityTrait].getName}.
+                     |Passing an object NOT extending ${classOf[SerializabilityTrait].getSimpleName} as a message may cause Akka to fall back to Java serialization during runtime.
```
PawelLipski, line 45: Actually, this message needs to be refined to distinguish between the case of message/event/state... pls open an issue for that

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  override def newPhase(prev: Phase): Phase =
+    new StdPhase(prev) {
+      override def apply(unit: global.CompilationUnit): Unit = {
+        val body = unit.body
+        val reporter = CrossVersionReporter(global)
+
+        val genericsNames = Seq(
+          "akka.actor.typed.Behavior",
+          "akka.persistence.typed.scaladsl.ReplyEffect",
```
PawelLipski, line 21: Pls add `Effect` here as well

PawelLipski: And open an issue for adding more types (I suppose we should check a few dozen different types to make this plugin really production-grade)

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,64 @@
...
+          "akka.projection.eventsourced.EventEnvelope")
+
+        annotatedTraitsCache = body
+          .collect {
+            case x: TypeTree if genericsNames.contains(x.tpe.typeSymbol.fullName) => x.tpe.typeArgs
+          }
+          .flatten
+          .foldRight(annotatedTraitsCache) { (next, annotatedTraits) =>
+            if (annotatedTraits.exists(next <:< _) || next.typeSymbol.fullName.startsWith("akka")) {
```
PawelLipski, line 30: Nit: maybe `startsWith("akka.")` ?

project/Dependencies.scala
```diff
@@ -32,4 +38,12 @@ object Dependencies {
 
   val borerDeps = Seq(borerCore, borerDerivation, borerAkka)
 
+  val scalaCompiler = "org.scala-lang" % "scala-compiler"
+  val scalaLibrary = "org.scala-lang" % "scala-library"
+  val scalaReflect = "org.scala-lang" % "scala-reflect"
+
+  private val scalaCompilerDeps = Seq(scalaCompiler, scalaLibrary, scalaReflect)
```
PawelLipski, line 45: Nit: `scalaPluginDeps`? This one and vals below

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
PawelLipski: (coz as for now, even if it event or state class that's not serializable, we'll still get an error about a "message")

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
MarconZet: > Pls add `Effect` here as well

Could you provide a FQCN of `Effect`?

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,64 @@
...
+          "akka.actor.typed.Behavior",
+          "akka.persistence.typed.scaladsl.ReplyEffect",
+          "akka.projection.eventsourced.EventEnvelope")
+
+        annotatedTraitsCache = body
+          .collect {
+            case x: TypeTree if genericsNames.contains(x.tpe.typeSymbol.fullName) => x.tpe.typeArgs
+          }
+          .flatten
+          .foldRight(annotatedTraitsCache) { (next, annotatedTraits) =>
```
PawelLipski, line 29: Hmmm does it need to be `foldRight`? Isn't it just a `filter` or sth like that?

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,64 @@
...
+                  if (pluginOptions.verbose)
+                    reporter.echo(
+                      s"${classOf[SaferSerializerPlugin].getSimpleName}: Found new annotated trait: ${annotatedType.typeSymbol.fullName}")
+                  annotatedType :: annotatedTraits
+                case None =>
+                  reporter.error(
+                    next.typeSymbol.pos,
+                    s"""${next.toString()} is used as Akka message but does not extend a trait annotated with ${classOf[
+                      SerializabilityTrait].getName}.
+                     |Annotate this class or one of the traits/classes it extends with @${classOf[SerializabilityTrait].getName}.
```
PawelLipski, line 44: Nit: swap this line with the next; it's more natural in error messages for the action point (line 44 now) to be at the very end, after the rationale of the error (line 45 now)

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
MarconZet: With fold right it updates cache during the fold

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
PawelLipski: Sorry, I meant `akka.persistence.typed.scaladsl.Effect`

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,64 @@
...
+
+import scala.tools.nsc.plugins.PluginComponent
+import scala.tools.nsc.{Global, Phase}
+
+class SaferSerializerPluginComponent(val pluginOptions: PluginOptions, val global: Global) extends PluginComponent {
+  import global._
+  override val phaseName: String = "akka-safer-serializer-gather"
+  override val runsAfter: List[String] = List("refchecks")
+
+  var annotatedTraitsCache: List[Type] = List()
```
PawelLipski, line 11: Async paranoia: any risk this collection might be modified concurrently by two threads? esp. when compiler is run with `-Ybackend-parallelism` > 1?

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SerializerTrait.scala
PawelLipski: Okay, seems resolved now, not sure which change fixed that :)

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
```diff
@@ -0,0 +1,64 @@
...
+              annotatedTraits
+            } else {
+              findSuperclassAnnotatedWithSerializabilityTrait(next) match {
+                case Some(annotatedType) =>
+                  if (pluginOptions.verbose)
+                    reporter.echo(
+                      s"${classOf[SaferSerializerPlugin].getSimpleName}: Found new annotated trait: ${annotatedType.typeSymbol.fullName}")
+                  annotatedType :: annotatedTraits
+                case None =>
+                  reporter.error(
```
PawelLipski, line 40: Hmm still getting false positives here (almost exclusively on upper bounds of type params)... details on Slack since they contain pieces of proprietary code

build.sbt
MarconZet: #33

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
MarconZet: According to [this article](https://kubuszok.com/2018/speed-up-things-in-scalac-and-sbt/) `-Ybackend-parallelism` is responsible for byte code generation.
And even if, a race condition in our scenario is of no consequence.

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
MarconZet: #34

plugin/src/main/scala/org/virtuslab/akkasaferserializer/SaferSerializerPluginComponent.scala
PawelLipski: Okay, good to know, I thought more stuff is parallelized with this flag

build.sbt
```diff
@@ -43,4 +42,25 @@ lazy val codecs = (projectMatrix in file("codecs"))
   .settings(libraryDependencies ++= borerDeps)
   .jvmPlatform(scalaVersions = supportedScalaVersions)
 
-// See https://www.scala-sbt.org/1.x/docs/Using-Sonatype.html for instructions on how to publish to Sonatype.
+lazy val pluginLibrary = (projectMatrix in file("plugin-library"))
+  .settings(name := "safer-serializer-library")
```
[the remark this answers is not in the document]
PawelLipski: So how about hmmm `akka-serializability-checker-{library,plugin}`? or sth like that?

checker-plugin/src/main/resources/scalac-plugin.xml
```diff
@@ -0,0 +1,4 @@
+<plugin>
+    <name>safer-serializer-plugin</name>
```
PawelLipski, line 2: @MarconZet at your convenience, pls change the plugin name here as well

checker-plugin/src/main/resources/scalac-plugin.xml
```diff
@@ -0,0 +1,4 @@
...
+    <classname>org.virtuslab.akkasaferserializer.SaferSerializerPlugin</classname>
```
PawelLipski, line 3: Aaaand class name

PR #46, opened 2021-06-16
Updating README.md

MarconZet:

PR #49, opened 2021-06-18
Project rename

MarconZet:


.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ jobs:
       # https://circleci.com/docs/2.0/configuration-reference/#available-machine-images
       image: ubuntu-2004:202101-01
 
-    working_directory: ~/akka-safer-serializer
```
PawelLipski, line 8: Looks like the only file with occurrences of `git grep -i safer` left is README.md

MarconZet: Yes, I already rebased #46

borer-akka-serializer/src/main/scala-2.12/org/virtuslab/akkaserializationhelper/RuntimeReflections.scala
```diff
@@ -1,4 +1,4 @@
-package org.virtuslab.akkasaferserializer
+package org.virtuslab.akkaserializationhelper
```
PawelLipski, line 1: Just thinking whether this should be called `org.virtuslab.ash` in package names 🤔

MarconZet: I agree

borer-akka-serializer/src/main/scala-2.12/org/virtuslab/ash/RuntimeReflections.scala
```diff
@@ -1,4 +1,4 @@
-package org.virtuslab.akkasaferserializer
+package org.virtuslab.ash
```
PawelLipski, line 1: I checked, it seems there aren't prominent projects called `ash` in Scala open source... this is the closest I could find: https://github.com/shadaj/ash but it's completely abandoned

PawelLipski (approved):
Approving, subject to the rename of the entire repo

PR #54, opened 2021-06-23
Hot-fix

### LINKED ISSUE #53: `dumpEventSchemaOutputFilename` does NOT allow for specifying absolute path
The output file is always assumed to live under `<subproject-directory>/target` :/

E.g. when I passed full path (`/home/plipski/hydra-backend/target/events-current.json`) to the file, I've got:

```
(hydra-backend / dumpEventSchema) java.nio.file.NoSuchFileException: /home/plipski/hydra-backend/services/target/home/plipski/hydra-backend/target/events-current.json
```

:/

MarconZet:
closes #53

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,4 +32,4 @@ jobs:
           command: sbt test
       - run:
           name: Testing sbt plugin
-          command: sbt "dumpEventSchemaCompilerPlugin / publishLocal" scripted
+          command: sbt clean "dumpEventSchemaCompilerPlugin / publishLocal" scripted
--- a/build.sbt
+++ b/build.sbt
@@ -19,19 +19,19 @@ Global / onChangedBuildSource := ReloadOnSourceChanges
 lazy val commonSettings = Seq(
   scalafmtOnCompile := true,
   scalacOptions ++= Seq(
-      "-deprecation",
-      "-encoding",
-      "UTF-8",
-      "-feature",
-      "-language:_",
-      "-Xfatal-warnings",
-      "-Xlog-reflective-calls",
-      "-Xlint:_",
-      "-Ybackend-parallelism",
-      "8",
-      "-Ywarn-dead-code",
-      "-Ywarn-unused:-imports,_",
-      "-unchecked"),
+    "-deprecation",
+    "-encoding",
+    "UTF-8",
+    "-feature",
+    "-language:_",
+    "-Xfatal-warnings",
+    "-Xlog-reflective-calls",
+    "-Xlint:_",
+    "-Ybackend-parallelism",
+    "8",
+    "-Ywarn-dead-code",
+    "-Ywarn-unused:-imports,_",
+    "-unchecked"),
   libraryDependencies ++= commonDeps)
 
 publish / skip := true
@@ -41,14 +41,14 @@ lazy val borerAkkaSerializer = (projectMatrix in file("borer-akka-serializer"))
   .settings(commonSettings)
   .settings(
     libraryDependencies ++= Seq(
-        borerCore,
-        akkaTyped % Provided,
-        reflections,
-        borerDerivation % Test,
-        akkaTestKit % Test,
-        akkaStream % Test,
-        akkaStreamTestKit % Test,
-        enumeratum % Test))
+      borerCore,
+      akkaTyped % Provided,
+      reflections,
+      borerDerivation % Test,
+      akkaTestKit % Test,
+      akkaStream % Test,
+      akkaStreamTestKit % Test,
+      enumeratum % Test))
   .dependsOn(borerExtraCodecs % Test)
   .jvmPlatform(scalaVersions = supportedScalaVersions)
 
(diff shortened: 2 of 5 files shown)


sbt-dump-event-schema/src/main/scala/org/virtuslab/ash/DumpEventSchema.scala
```diff
@@ -24,7 +24,8 @@ object DumpEventSchema extends DefaultJsonProtocol {
     }
 
     for {
-      writer <- outputFile.clear().newFileOutputStream().printWriter().autoClosed
+      writer <-
```
PawelLipski, line 27: Whoops... I've tried this implementation, and even when I specify an absolute path, the target file is located under:

`./services/target/home/plipski/hydra-backend/target/foo.json`

instead of:

`/home/plipski/hydra-backend/target/foo.json`

:/

PawelLipski: So it looks like the current impl treats absolute paths as relative ones :/

MarconZet:
@PawelLipski I added two new keys: `dumpEventSchemaOutputDirectoryPath` and `dumpEventSchemaOutputFilePath`. You want to use the first one.

sbt-dump-event-schema/src/main/scala/org/virtuslab/ash/DumpEventSchemaKeys.scala
```diff
@@ -8,7 +8,9 @@ import sbt.taskKey
 trait DumpEventSchemaKeys {
   lazy val dumpEventSchema = taskKey[Unit]("Dump event schema")
   lazy val dumpEventSchemaOutputFile = taskKey[File]("Output file to dump event schema to")
+  lazy val dumpEventSchemaOutputFilePath = settingKey[String]("Path to event schema dump file")
   lazy val dumpEventSchemaOutputFilename = settingKey[String]("Filename to dump event schema to")
+  lazy val dumpEventSchemaOutputDirectoryPath = settingKey[String]("Directory path to event schema dump")
```
PawelLipski, line 13: Are all 4 keys still necessary? Esp. the one `dumpEventSchemaOutputFile` that isn't likely (?) to be ever directly set by the users?

PawelLipski: I.e. a more general question, does **sbt's project model use keys** also for the **internal properties** of the plugin? Or are the keys considered to be more or less the "interface" of a plugin?

.circleci/config.yml
```diff
@@ -32,4 +32,4 @@ jobs:
           command: sbt test
       - run:
           name: Testing sbt plugin
-          command: sbt "dumpEventSchemaCompilerPlugin / publishLocal" scripted
+          command: sbt clean "dumpEventSchemaCompilerPlugin / publishLocal" scripted
```
PawelLipski, line 35: Hmm is this `clean` necessary here? If so, pls drop a comment as to why

build.sbt
```diff
@@ -101,10 +101,11 @@ lazy val dumpEventSchema = (project in file("sbt-dump-event-schema"))
     publishM2Configuration := publishM2Configuration.value.withResolverName(localMavenResolverForSbtPlugins.name),
     scriptedLaunchOpts := {
       scriptedLaunchOpts.value ++
-      Seq("-Xmx1024M", "-Dplugin.version=" + version.value)
+        Seq("-Xmx1024M", "-Dplugin.version=" + version.value)
     },
     scriptedBufferLog := false)
 
+//Warning: when publishing artefacts two times, without clean in between, sometimes instead of jar tar, plugin is published as normal jar
```
PawelLipski, line 108: Maybe open an issue and reference and here and in config.yml

Also, for:
1. easy searching of places where the given issue is referenced
2. to make https://virtuslab.github.io/tips/#jetbrains-products/linking-issues-in-jb-ides possible
3. to prevent dangling TODOx/FIXMEs with no associated github issue,

pls add a script like this one: https://github.com/VirtusLab/git-machete-intellij-plugin/blob/develop/scripts/enforce-issue-number-for-todos and run from the CI

PawelLipski: (can be on a separate PR ofc)

PawelLipski (approved):
Approving, subject to opening an issue on non-reproducible builds

sbt-dump-event-schema/src/main/scala/org/virtuslab/ash/DumpEventSchemaKeys.scala
MarconZet: Yes but no.
[This plugin](https://github.com/sbt/sbt-assembly/blob/9a2ca181226a8b437a784792343a203a0b94a6b8/src/main/scala/sbtassembly/AssemblyPlugin.scala#L76) is written by sbt creator. Interesting options are: `assemblyOutputPath` (actually a file) and `assemblyJarName`, which is used in creating `assemblyOutputPath`

Issue #24, opened 2021-05-04
Publish to Maven Central


PawelLipski:
- [ ] Add a badge from shields.io to README
- [ ] Add code to CI pipeline that would do a release on `master` builds, and preferably publish a snapshot build on `develop` builds as well

PR #55, opened 2021-06-25
Jitpack fix

MarconZet:
I have no idea why, but scalafix interferes with Jitpack, further inquiry needed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,9 +18,9 @@ jobs:
       - run:
           name: Checking formating
           command: sbt scalafmtCheckAll
-      - run:
-          name: Checking import order
-          command: sbt "scalafixAll --check"
+#      - run: TODO: uncomment when #24 is done
+#          name: Checking import order
+#          command: sbt "scalafixAll --check"
       - run:
           name: Compiling production code
           command: sbt compile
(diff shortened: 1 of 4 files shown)


PawelLipski:
Hope it gets resolved with migration to Maven Central 😅

.circleci/config.yml
```diff
@@ -18,9 +18,9 @@ jobs:
       - run:
           name: Checking formating
           command: sbt scalafmtCheckAll
-      - run:
-          name: Checking import order
-          command: sbt "scalafixAll --check"
+#      - run:
```
PawelLipski, line 21: Pls add a TODO for #24 here so that we know when this can be reverted

jitpack.yml
```diff
@@ -0,0 +1,2 @@
+jdk:
+  - openjdk16
```
PawelLipski, line 2: Okay... does it matter now that scalafix is disabled anyway?

PawelLipski (approved):

jitpack.yml
MarconZet: OpenJDK16 has better metaspace management, so it may help us delay the problem even futher.
See [here](https://openjdk.java.net/jeps/387)

PR #64, opened 2021-07-09
Added Borer and Circe migration examples

MarconZet:


examples/borer-circe-migration/build.sbt
```diff
@@ -0,0 +1,23 @@
+name := "borer-circe-migration"
+version := "0.1"
+scalaVersion := "2.13.6"
+
+idePackagePrefix := Some("pl.marconzet")
```
PawelLipski, line 5: Hmm maybe use `com.example` package instead... anyway, I assume https://marconzet.pl/ is your domain? ;)

PawelLipski: Or even `org.virtuslab.example`, usually I see that projects keep the examples in the similar packages as the actual production code

examples/borer-circe-migration/build.sbt
```diff
@@ -0,0 +1,23 @@
+name := "borer-circe-migration"
```
PawelLipski, line 1: Hmmm so `examples/` consistute a completely separate sbt project? AFAICS it's not being built together with the main project, right?

PawelLipski: Is there any specific reason for that? I'd rather expect the examples to be a part of the main project, just never published

examples/borer-circe-migration/project/plugins.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("org.jetbrains" % "sbt-ide-settings" % "1.1.0")
```
PawelLipski, line 1: Hmm does it make sense to add it to the sbt plugins of the root project?

PawelLipski: It's only used for `idePackagePrefix`, right?

examples/borer-circe-migration/src/main/scala/btest/Test1.scala
```diff
@@ -0,0 +1,25 @@
+package pl.marconzet
```
PawelLipski, line 1: Hmm `pl/marconzet` folders are missing from path... is it intentional? 🤔

examples/borer-circe-migration/src/main/scala/btest/Test1.scala
```diff
@@ -0,0 +1,25 @@
...
+package btest
```
PawelLipski, line 2: What does `b` stand for here?

PawelLipski: Oh ok, borer-test vs circe-test... I'd still use more verbatim names

examples/borer-circe-migration/src/main/scala/btest/Test1.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import io.bullet.borer.{Codec, Json}
+import io.bullet.borer.derivation.ArrayBasedCodecs.{deriveAllCodecs, deriveCodec}
+
+object Test1 extends App {
```
PawelLipski, line 7: Hmm instead of `Test1`... maybe `BasicCodecs`? or anyway, sth that explain the purpose more clearly

PawelLipski: Also, do we even need this example? since no migration is involved here, it's a Borer-only piece

examples/borer-circe-migration/src/main/scala/btest/Transcoding.scala
```diff
@@ -0,0 +1,46 @@
+package pl.marconzet
```
PawelLipski, line 1: Also... the folder name `borer-circe-migration` is heavily misleading, it suggests that some stuff is migrated **between Borer and Circe** (which wouldn't be of much value)

PawelLipski: Maybe just `examples/event-migration/`? and package structure will then show which libraries are considered (also, what if examples for a third library are added... `borer-circe-avro-migration`? ;) )

examples/borer-circe-migration/src/main/scala/btest/Transcoding.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  case class Employee2(nick: String, age: Option[Int])
+
+  implicit val employee1Codec: Codec[Employee1] = deriveCodec[Employee1]
+  implicit val employee2Codec: Codec[Employee2] = deriveCodec[Employee2]
+
+  val value = Employee2("Alice", None)
+
+
+  val dom = Json // could also be `Json` to target JSON
```
PawelLipski, line 19: ❓ it's already `Json` ;)

examples/borer-circe-migration/src/main/scala/btest/Transcoding.scala
```diff
@@ -0,0 +1,46 @@
...
+    .transEncode(value)
+    .transDecode
+    .to[Dom.Element]
+    .value
+
+  val transformer =
+    new Dom.Transformer {
```
PawelLipski, line 26: `Dom` stands for Document Object Model here? 😯

PawelLipski: As in, I'd say Document Object Model is only limited to XML/HTML... of course a matter of nomenclature, since logically it's a pretty similar thing anyway here

examples/borer-circe-migration/src/main/scala/btest/Transcoding.scala
```diff
@@ -0,0 +1,46 @@
...
+      override def transformMapMember(member: (Element, Element)): (Element, Element) =
+        member match {
+          case (k@StringElem("age"), ArrayElem.Unsized(elements)) => k -> elements.headOption.getOrElse(IntElem(20))
+          case x => super.transformMapMember(x)
+        }
+    }
+
+  val transformed = transformer(dom)
+
+  val result = Json // could also be `Json` to target JSON
```
PawelLipski, line 39: Ditto, the comment is off

examples/borer-circe-migration/src/main/scala/btest/Transcoding.scala
```diff
@@ -0,0 +1,46 @@
...
+
+object Transcoding extends App {
+  case class Employee1(nick: String, age: Int)
+
+  case class Employee2(nick: String, age: Option[Int])
+
+  implicit val employee1Codec: Codec[Employee1] = deriveCodec[Employee1]
+  implicit val employee2Codec: Codec[Employee2] = deriveCodec[Employee2]
+
+  val value = Employee2("Alice", None)
```
PawelLipski, line 16: Hmmm okay so this naming here suggests that you're transcoding from an **newer model to an older one**, right? Will this be the more common scenario, or rather will transcoding from **older to newer** model be more common?

examples/borer-circe-migration/src/main/scala/btest/TranscodingCreateOptional.scala
```diff
@@ -0,0 +1,36 @@
+package pl.marconzet
```
PawelLipski, line 1: Lemme proceed with further sources once the existing comments are resolved since I guess some stuff is gonna change around here...

examples/borer-circe-migration/build.sbt
MarconZet: > I'd rather expect the examples to be a part of the main project

From my experience, on the contrary. Example being [grpc-java](https://github.com/grpc/grpc-java) where [examples](https://github.com/grpc/grpc-java/tree/master/examples) are a separate project.

I don't see any merit in putting the examples in a main build tool, only drawbacks, including:
1. Examples are not self contained, they start depending on files outside of them
2. Figuring out simple dependencies requires analysing big buildfile
3. Clutters main build file with stuff we don't care about

examples/borer-circe-migration/project/plugins.sbt
MarconZet: Calling it *only* is an understatement

MarconZet: > Hmm does it make sense to add it to the sbt plugins of the root project?

I tried, it doesn't work well with project matrix.

examples/borer-circe-migration/src/main/scala/btest/Test1.scala
MarconZet: `idePackagePrefix`

examples/borer-circe-migration/src/main/scala/btest/Transcoding.scala
MarconZet: It's [Document Object Model](https://sirthias.github.io/borer/borer-core/DOM.html)

examples/borer-circe-migration/build.sbt
PawelLipski: Ok, makes sense 👍🏻 
Either way, are the examples now being compiled within CI pipeline?

examples/borer-circe-migration/src/main/scala/btest/Test1.scala
PawelLipski: Ok... but we only want to apply that plugin to examples and not to the entire project?

examples/borer-circe-migration/project/plugins.sbt
PawelLipski: > Calling it only is an understatement

Ok, for clarity coz I haven't used it before &mdash; the effect of the plugin on the project is specifically that you can skip the folders (`org/virtuslab/example`) in examples module, right?

PawelLipski: Or are there any other benefits? I'd be concerned about making the examples confusing for the users who don't know `sbt-ide-settings`, let alone those who don't use IntelliJ (although the latter is rather a small minority)

MarconZet: > the effect of the plugin on the project is specifically that you can skip the folders
Yes, this is the main effect

MarconZet: Removing the plugin is not an issue. I agree that it may be confusing for some.

examples/borer-circe-migration/build.sbt
MarconZet: That would require two things:
* Rewriting examples to scalatest (would like do anyway)
* Putting new command in .circleci
Implementing it is not a problem, but I would like to wait with final decision wether we run it with CI until we see what kind of impact it will have on CI execution time.

examples/event-migration/src/test/scala/org/virtuslab/example/CirceMigrationExamples.scala
```diff
@@ -0,0 +1,84 @@
...
+    case class Employee1(nick: String)
+    case class Employee2(nick: String, age: Option[Int])
+
+    val json = Employee1("Alice").asJson.noSpaces
+    val result = decode[Employee2](json).getOrElse(throw new RuntimeException)
+    assert(result == Employee2("Alice", None))
+  }
+
+  test("Add required field") {
+    object Def {
```
MarconZet, line 21: There's some macro magic going on in the background that doesn't like ScalaTest

MarconZet: Sometimes requiring this workaround

examples/event-migration/src/test/scala/org/virtuslab/example/BorerMigrationExamples.scala
```diff
@@ -0,0 +1,128 @@
+package org.virtuslab.example
+
+import io.bullet.borer.derivation.MapBasedCodecs.deriveCodec
+import io.bullet.borer.{Codec, Dom, Json}
+import org.scalatest.funsuite.AnyFunSuiteLike
+
+class BorerMigrationExamples extends AnyFunSuiteLike {
```
MarconZet, line 7: New presentation of migration examples by using ScalaTest

examples/borer-circe-migration/project/plugins.sbt
PawelLipski: Yeah... I'd rather refrain from non-standard solutions in the examples :/

examples/borer-circe-migration/build.sbt
PawelLipski: Okay! pls give a try, I doubt it's gonna significantly increase the execution time, compared to what we've got now

PawelLipski: Rel, https://github.com/VirtusLab/akka-serialization-helper/issues/67

examples/event-migration/project/plugins.sbt
```diff
@@ -0,0 +1 @@
+
```
PawelLipski, line 1: ❓

examples/event-migration/build.sbt
```diff
@@ -0,0 +1,26 @@
+name := "event-migration"
+version := "0.1"
+scalaVersion := "2.13.6"
```
PawelLipski, line 3: Pls bump from 2.13.5 in the other places as well

examples/event-migration/build.sbt
```diff
@@ -0,0 +1,26 @@
...
+
+val circeVersion = "0.14.1"
+val borerVersion = "1.7.1"
```
PawelLipski, line 6: Hmmm 1.7.1 here vs 1.7.2 in the rest of the project?

examples/event-migration/build.sbt
```diff
@@ -0,0 +1,26 @@
...
+val scalaTestVersion = "3.2.9"
```
PawelLipski, line 7: Same, 3.2.2 vs 3.2.9?

examples/event-migration/build.sbt
PawelLipski: Theoretically doesn't need to be the same in examples and in the rest of the project... but still

examples/event-migration/build.sbt
```diff
@@ -0,0 +1,26 @@
...
+
+libraryDependencies ++= Seq(
+  "io.bullet" %% "borer-core",
+  "io.bullet" %% "borer-derivation",
+  "io.bullet" %% "borer-compat-akka",
+  "io.bullet" %% "borer-compat-circe",
+  "io.bullet" %% "borer-compat-scodec").map(_ % borerVersion)
+
+libraryDependencies ++= Seq(
+  "org.scalactic" %% "scalactic" % scalaTestVersion,
```
PawelLipski, line 23: Hmmm this is in `Compile` and not `Test` scope? 🤔

examples/event-migration/src/test/scala/org/virtuslab/example/CirceMigrationExamples.scala
```diff
@@ -0,0 +1,84 @@
...
+  test("Remove field") {
+    case class Employee1(nick: String, age: Int)
+    case class Employee2(nick: String)
+
+    val json = Employee1("Alice", 10).asJson.noSpaces
+    val result = decode[Employee2](json).getOrElse(throw new RuntimeException)
+    assert(result == Employee2("Alice"))
+  }
+
+  test("Rename filed") {
```
PawelLipski, line 43: `field`

examples/borer-circe-migration/build.sbt
MarconZet: 1 minute - not great, not bad

examples/event-migration/build.sbt
MarconZet: #70

examples/event-migration/build.sbt
MarconZet: #70

PawelLipski (approved):

PR #65, opened 2021-07-12
Fixed dump-event-schema-compiler-plugin non-reproducible builds

### LINKED ISSUE #61: dump-event-schema-compiler-plugin non-reproducible builds
When publishing artefacts two times, without clean in between, sometimes instead of jar tar, plugin is published as normal jar

MarconZet:
closes #61

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -103,9 +103,17 @@ lazy val dumpEventSchema = (project in file("sbt-dump-event-schema"))
       scriptedLaunchOpts.value ++
       Seq("-Xmx1024M", "-Dplugin.version=" + version.value)
     },
+    scriptedDependencies := { // publishing compiler plugin locally for testing
+      scriptedDependencies.value
+      (dumpEventSchemaCompilerPlugin
+        .filterProjects(Seq(ScalaVersionAxis(scalaVersion212, "2.12")))
+        .head / publishLocal).value
+      (dumpEventSchemaCompilerPlugin
+        .filterProjects(Seq(ScalaVersionAxis(scalaVersion213, "2.13")))
+        .head / publishLocal).value
+    },
     scriptedBufferLog := false)
 
-//Warning: when publishing artefacts two times, without clean in between, sometimes instead of jar tar, plugin is published as normal jar
 lazy val dumpEventSchemaCompilerPlugin = (projectMatrix in file("dump-event-schema-compiler-plugin"))
   .enablePlugins(AssemblyPlugin)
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -103,9 +103,17 @@ lazy val dumpEventSchema = (project in file("sbt-dump-event-schema"))
       scriptedLaunchOpts.value ++
       Seq("-Xmx1024M", "-Dplugin.version=" + version.value)
     },
+    scriptedDependencies := { //publishing locally compiler plugin for testing
```
PawelLipski, line 106: Nit:
```suggestion
    scriptedDependencies := { // publishing compiler plugin locally for testing
```

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.5.2
+sbt.version=1.5.4
```
PawelLipski, line 1: Nit: would it make sense/be feasible to share this version b/w all 3 `build.properties` files somehow?

PawelLipski: Ofc symlink being the crudest way of doing that 😅

build.sbt
```diff
@@ -121,6 +129,8 @@ lazy val dumpEventSchemaCompilerPlugin = (projectMatrix in file("dump-event-sche
         .getOrElse(Seq.empty)
     },
     libraryDependencies ++= Seq(sprayJson, betterFiles, akkaTyped % Test, akkaPersistence % Test),
+    packageBin / publishArtifact := false, //we want to publish fat jar
+    Compile / packageBin / artifactPath := crossTarget.value / "packageBinPlaceholder.jar", //this ensures that normal jar doesn't override fat jar
```
PawelLipski, line 133: Hmmm so what exactly was the problem happening here? was it some kind of race condition? 🤔

MarconZet: Yes, it was race condition. The artefact that took longer to create was the one being published. Running `clean` before `publishM2` ensured that the right artefact won the race.

PawelLipski: Ok, if you could add a mention of the possible race condition to the code so that we're wary of avoiding a regression in the future

project/build.properties
MarconZet: Symlink won't work, because sbt copies project to temporary directory for testing

MarconZet: On the other hand, I managed to merge to tests into one, using cross building

PawelLipski (approved):
Approving, subject to the comment on possible regression

PR #77, opened 2021-08-20
Compiler plugin for checking registration of serializable types

### LINKED ISSUE #57: Add a checker for serializers like Borer/Circe/Kryo to make sure that all top-level traits are explicitly registered

MarconZet:
closes #57


registration-checker-compiler-plugin/src/main/resources/scalac-plugin.xml
```diff
@@ -0,0 +1,4 @@
+<plugin>
+    <name>dump-event-schema-plugin</name>
```
PawelLipski, line 2: Whoops

serializability-checker-library/src/main/scala/org/virtuslab/ash/Serializer.scala
```diff
@@ -0,0 +1,6 @@
+package org.virtuslab.ash
+
+import scala.annotation.nowarn
+
+@nowarn("cat=unused")
```
PawelLipski, line 5: Whoops what's unused here? 🤔

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/RegistrationCheckerCompilerPlugin.scala
```diff
@@ -0,0 +1,59 @@
...
+import java.nio.file.{Files, NoSuchFileException}
+import java.util.function.Consumer
+import scala.collection.mutable
+import scala.tools.nsc.Global
+import scala.tools.nsc.plugins.{Plugin, PluginComponent}
+
+class RegistrationCheckerCompilerPlugin(override val global: Global) extends Plugin {
+  override val name: String = "registration-checker-plugin"
+  override val description: String =
+    "checks whether classes marked with serializability trait are being referenced in a marked serializer"
```
PawelLipski, line 15: Could you explain why there are two phases here? Is it just matter of code style, or is there a technical requirement for that?

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/ClassSweepCompilerPluginComponent.scala
```diff
@@ -0,0 +1,42 @@
...
+import scala.collection.mutable
+import scala.tools.nsc.{Global, Phase}
+import scala.tools.nsc.plugins.PluginComponent
+
+class ClassSweepCompilerPluginComponent(options: RegistrationCheckerOptions, override val global: Global)
+    extends PluginComponent {
+  import global._
+  override val phaseName: String = classSweepPhaseName
+  override val runsAfter: List[String] = List("refchecks")
+  override def description: String = s"searches for direct descendants of classes annotated with serializability trait"
```
PawelLipski, line 14: Hmmm why do we care about _direct descendants_? Don't we check just for the annotated classes itself, whether they're referenced from the serializer class?

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/RegistrationCheckerCompilerPlugin.scala
MarconZet: Using two phases guarantees correct ordering

serializability-checker-library/src/main/scala/org/virtuslab/ash/Serializer.scala
MarconZet: `clazz` field

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/RegistrationCheckerCompilerPlugin.scala
PawelLipski: Hmmm okay... but wouldn't the same be achieved if searching for class AND THEN checking serializer were placed one after another in the same compilation phase?

MarconZet: That would require either ordering source files or knowing that the current source file is the last one. We lack the necessary API  to make it work.

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/ClassSweepCompilerPluginComponent.scala
MarconZet: That way the only class that is annotated, is the serializability trait itself.
They must be direct descendants, because we need to know how to serialize all named classes of serializability trait.

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/RegistrationCheckerCompilerPlugin.scala
```diff
@@ -0,0 +1,61 @@
...
+      case None =>
+        error("No directory for saving cache file specified")
+        false
+    }
+  }
+  override val optionsHelp: Option[String] = Some("directory where cache file will be saved")
+}
+
+object RegistrationCheckerCompilerPlugin {
+  val classSweepPhaseName = "tmp1"
```
PawelLipski, line 58: Whoops

PawelLipski (approved):
Approving, subject to fixing compiler phase names

Issue #26, opened 2021-05-04
Write a README


PawelLipski:
- [x] CI + License + Maven badge (shields.io, see https://github.com/VirtusLab/git-machete-intellij-plugin/ for a sample)
- [x] Logo (#25)
- [x] Build setup (compiler plugin + dependency)
- [x] Rationale: why Jackson sucks for Akka and how it's done better here
- [x] Remember about explicit mentions of Borer! so that it doesn't create an impression that we attribute to ourselves what's actually done by Borer
- [x] Document sbt settings for dump event schema plugin/options for compiler plugin (maybe not in README but in a separate file? TBD)
- [x] Document Circe serializer with some Scala snippets

MarconZet:
3, 4 and 5 done by #39

PR #80, opened 2021-08-30
Added disable feature for registration checker plugin

MarconZet:


registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/RegistrationCheckerCompilerPlugin.scala
```diff
@@ -20,6 +20,9 @@ class RegistrationCheckerCompilerPlugin(override val global: Global) extends Plu
   override val components: List[PluginComponent] = List(classSweep, serializerCheck)
 
   override def init(options: List[String], error: String => Unit): Boolean = {
+    if (options.contains("--disable"))
```
PawelLipski, line 23: Actually... the option is only applicable to this compiler plugin but not to to class sweeper? 🤔

MarconZet: This option disables all of the compiler plugin, which means `PluginComponents` are never consumed by the compiler and new phases are never created.

MarconZet: From `def init` documentation:

> The plugin can opt out of further processing by returning false.  For example, if the plugin has an "enable" flag, now would be a good time to sit on the bench.

PawelLipski: Aww okay I forgot there is single compiler plugin with two components (rather than two compiler plugins) 🤦🏻 nvm

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/RegistrationCheckerCompilerPlugin.scala
```diff
@@ -51,7 +54,10 @@ class RegistrationCheckerCompilerPlugin(override val global: Global) extends Plu
         false
     }
   }
-  override val optionsHelp: Option[String] = Some("directory where cache file will be saved")
+  override val optionsHelp: Option[String] = Some("""
+      |. - directory where cache file will be saved, required
+      |--disable - disables the plugin
```
PawelLipski, line 59: While you're at this change... could you add the relevant section about this compiler plugin to README.md?

MarconZet: I prefer to do it as part of #26, in the later PR.

PawelLipski: Okay, added a checkbox there so that it's not lost

PawelLipski (approved):

PR #81, opened 2021-08-31
Release to maven central using sbt-ci-release plugin and GitHub Actions

MarconZet:


.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,16 @@
+name: Release Drafter
```
PawelLipski, line 1: Actually... we've got both CircleCI and GH actions now... let's make it consistent

PawelLipski: AFAICS there's quite a lot of ready logic in these 2 actions that might not be readily available in CircleCI... so maybe the existing CircleCI pipeline (which is rather trivial) could be migrated to a GH action

PawelLipski: Oh okay, mostly `release-drafter` since the other one has just `setup-scala`

project/plugins.sbt
```diff
@@ -1,8 +1,5 @@
 addSbtPlugin("com.eed3si9n" % "sbt-projectmatrix" % "0.8.0")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "1.0.0")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.2")
-/**
- * Adding this plugin to sbt causes `java.lang.OutOfMemoryError: Metaspace` when publishing artifacts with Jitpack.
- * The plugin itself doesn't do anything bad, but adding it was the final straw for limited memory space.
- */
-//addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.29")
\ No newline at end of file
+//addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.30")
```
PawelLipski, line 4: Uncomment now? ;)

build.sbt
```diff
@@ -1,15 +1,22 @@
...
-import sbt.io.Path.userHome
-import sbt.librarymanagement.Patterns
 
 lazy val supportedScalaVersions = List(scalaVersion213, scalaVersion212)
 
 ThisBuild / scalaVersion := supportedScalaVersions.head
-ThisBuild / version := "0.1.0-SNAPSHOT"
 ThisBuild / organization := "org.virtuslab"
 ThisBuild / organizationName := "VirtusLab"
+ThisBuild / versionScheme := Some("early-semver")
```
PawelLipski, line 11: `early` as in 0.x.y rather than (n>=1).x.y?

.github/workflows/release-drafter.yml
MarconZet: opened #84

build.sbt
MarconZet: Enumeration of version schemes [here](https://www.scala-sbt.org/1.x/docs/Publishing.html#Version+scheme)

.circleci/config.yml
```diff
@@ -18,15 +18,15 @@ jobs:
...
+      - run:
+          name: Checking import order
+          command: sbt "scalafixAll --check"
       - run:
           name: Compiling production code
           command: sbt compile
       - run:
           name: Compiling tests
-          command: sbt test:compile
+          command: sbt Test/compile
```
PawelLipski, line 29: sbt CLI/REPL is madness 🤯 I've got no idea why this was needed but I trust you it was 😅

.github/workflows/release.yml
```diff
@@ -0,0 +1,19 @@
...
+  publish:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 0
+      - uses: olafurpg/setup-scala@v10
+      - run: sbt ci-release
+        env:
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
```
PawelLipski, line 16: Hmmm actually... are all these 4 secrets some VirtusLab-wide creds? how have you gotten them, is there some LastPass that I can have access to as well?

.github/workflows/release.yml
```diff
@@ -0,0 +1,19 @@
+name: Release
+on:
+  push:
+    branches: [main]
+    tags: ["*"]
```
PawelLipski, line 5: Paranoia/sanity check, is there any scenario given the current GH repo settings where someone could extract these secrets by opening a PR from a fork? (secrets are probably masked in GH actions output but only if matching 1:1... there'd be no problem e.g. base64-encoding them first)

PawelLipski: (In CircleCI, I think we've got `Pass secrets to builds from forks` disabled)

.github/workflows/release.yml
MarconZet: I got these secrets form forwarded emails

.github/workflows/release.yml
MarconZet: > Secrets are not passed to workflows that are triggered by a pull request from a fork.

.github/workflows/release.yml
PawelLipski: image

PawelLipski (approved):

PR #85, opened 2021-09-01
Added checks in CirceAkkaSerializer against NullPointerExceptions

### LINKED ISSUE #83: `org.virtuslab.ash.circe.CirceAkkaSerializer#scanForSerializables` should throw rather than log an error

MarconZet:
closes #83

### DIFF
--- a/circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
+++ b/circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
@@ -19,13 +19,18 @@ import org.reflections8.Reflections
 abstract class CirceAkkaSerializer[Ser <: AnyRef: ClassTag](system: ExtendedActorSystem)
     extends SerializerWithStringManifest
     with AkkaCodecs {
-  private val log = Logging(system, getClass)
-  private val conf = system.settings.config.getConfig("org.virtuslab.ash")
-  private val isDebugEnabled = conf.getBoolean("verbose-debug-logging") && log.isDebugEnabled
-
   val codecs: Seq[(ru.TypeTag[_ <: Ser], (Encoder[_ <: Ser], Decoder[_ <: Ser]))]
   val manifestMigrations: Seq[(String, String)]
   val packagePrefix: String
+
+  private val assertionMessage = " must be declared as a def or a lazy val to work correctly"
+  assert(codecs != null, "codecs" + assertionMessage)
+  assert(manifestMigrations != null, "manifestMigrations" + assertionMessage)
+  assert(packagePrefix != null, "packagePrefix" + assertionMessage)
+
+  private val log = Logging(system, getClass)
+  private val conf = system.settings.config.getConfig("org.virtuslab.ash")
+  private val isDebugEnabled = conf.getBoolean("verbose-debug-logging") && log.isDebugEnabled
 
   private val mirror = ru.runtimeMirror(getClass.getClassLoader)
   private val parents = codecs.flatMap { x =>
@@ -114,27 +119,30 @@ abstract class CirceAkkaSerializer[Ser <: AnyRef: ClassTag](system: ExtendedActo
     }
   }
 
-  private def scanForSerializables(packagePrefix: String): Unit = {
+  private def checkSerializableTypesForMissingCodec(packagePrefix: String): Unit = {
     val reflections = new Reflections(packagePrefix)
     val foundSerializables = reflections.getSubTypesOf(classTag[Ser].runtimeClass).asScala.filterNot(_.isInterface)
     foundSerializables.foreach { clazz =>
       try {
-        val codec = codecsMap(parents(clazz.getName))
-        assert(codec._1 != null)
-        assert(codec._2 != null)
+        codecsMap(parents(clazz.getName))
       } catch {
-        case e @ (_: NoSuchElementException | _: NotSerializableException) =>
+        case _: NoSuchElementException =>
           log.error(
-            e,
-            s"No codec found for [{}] class. Call Register[A] for this class or its supertype and append result to codecs.",
-            clazz.getName)
-        case e: AssertionError =>
-          log.error(
-            e,
-            "Codec for [{}] is null. If this codec is custom defined, use def or lazy val instead of val.",
+            s"No codec found for [{}] class. Call Register[A] for this class or its supertype and append the result to codecs.",
             clazz.getName)
       }
     }
   }
-  scanForSerializables(packagePrefix)
+
+  private def checkCodecsForNull(): Unit = {
+    codecs.foreach { registration =>
+      val (tag, (encoder, decoder)) = registration
+      if (encoder == null || decoder == null)
+        throw new AssertionError(
+          s"Codec for [${tag.tpe.typeSymbol.fullName}] is null. If this codec is custom defined, declare it as a def or lazy val instead of val.")
+    }
+  }
+
+  checkSerializableTypesForMissingCodec(packagePrefix)
+  checkCodecsForNull()
 }


circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
```diff
@@ -119,22 +124,26 @@ abstract class CirceAkkaSerializer[Ser <: AnyRef: ClassTag](system: ExtendedActo
...
+  private def checkCodecsForNull(): Unit = {
+    codecs.foreach { registration =>
+      val (tag, (encoder, decoder)) = registration
+      if (encoder == null || decoder == null)
+        throw new AssertionError(
+          s"Codec for [${tag.tpe.typeSymbol.fullName}] is null. If this codec is custom defined, use def or lazy val instead of val.")
+    }
+  }
+
   scanForSerializables(packagePrefix)
```
PawelLipski, line 147: For clarity, maybe `checkSerializableTypesForMissingCodec`?

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
```diff
@@ -119,22 +124,26 @@ abstract class CirceAkkaSerializer[Ser <: AnyRef: ClassTag](system: ExtendedActo
...
       }
     }
   }
+
+  private def checkCodecsForNull(): Unit = {
+    codecs.foreach { registration =>
+      val (tag, (encoder, decoder)) = registration
+      if (encoder == null || decoder == null)
+        throw new AssertionError(
+          s"Codec for [${tag.tpe.typeSymbol.fullName}] is null. If this codec is custom defined, use def or lazy val instead of val.")
```
PawelLipski, line 143:
```suggestion
          s"Codec for [${tag.tpe.typeSymbol.fullName}] is null. If this codec is custom defined, declare it as a def or lazy val instead of val.")
```

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
```diff
@@ -119,22 +124,26 @@ abstract class CirceAkkaSerializer[Ser <: AnyRef: ClassTag](system: ExtendedActo
...
       try {
-        val codec = codecsMap(parents(clazz.getName))
-        assert(codec._1 != null)
-        assert(codec._2 != null)
+        codecsMap(parents(clazz.getName))
       } catch {
         case e @ (_: NoSuchElementException | _: NotSerializableException) =>
           log.error(
             e,
             s"No codec found for [{}] class. Call Register[A] for this class or its supertype and append result to codecs.",
```
PawelLipski, line 132: :grammar-nazi:
```suggestion
            s"No codec found for [{}] class. Call Register[A] for this class or its supertype and append the result to codecs.",
```

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
```diff
@@ -19,14 +19,19 @@ import org.reflections8.Reflections
...
   val packagePrefix: String
 
+  private val assertionMessage = " must be declared as a def or a lazy val to work correctly"
+  assert(codecs != null, "codecs" + assertionMessage)
+  assert(manifestMigrations != null, "manifestMigrations" + assertionMessage)
+  assert(packagePrefix != null, "packagePrefix" + assertionMessage)
+
+  private val log = Logging(system, getClass)
+  private val conf = system.settings.config.getConfig("org.virtuslab.ash")
+  private val isDebugEnabled = conf.getBoolean("verbose-debug-logging") && log.isDebugEnabled
```
PawelLipski, line 33: This can't ever throw a `ConfigException.Missing` right? since the key is present in reference.conf

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
```diff
@@ -119,22 +124,26 @@ abstract class CirceAkkaSerializer[Ser <: AnyRef: ClassTag](system: ExtendedActo
     val foundSerializables = reflections.getSubTypesOf(classTag[Ser].runtimeClass).asScala.filterNot(_.isInterface)
     foundSerializables.foreach { clazz =>
       try {
-        val codec = codecsMap(parents(clazz.getName))
-        assert(codec._1 != null)
-        assert(codec._2 != null)
+        codecsMap(parents(clazz.getName))
       } catch {
         case e @ (_: NoSuchElementException | _: NotSerializableException) =>
```
PawelLipski, line 129: Hmmm how can a `NotSerializableException` be thrown from two `Map.apply` calls in the `try` block? (AFAICS)

PawelLipski: Also, is that exception even relevant here? it seems to correspond to Java serialization, right?

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
MarconZet: Yes, `akka-jackson-serializer` does the same thing.

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
MarconZet: > Hmmm how can a `NotSerializableException` be thrown from two `Map.apply` calls in the `try` block? (AFAICS)

Yeah, you are right, it can't

MarconZet: > Also, is that exception even relevant here? it seems to correspond to Java serialization, right?

This exception is very relevant for Akka serializers. From documentation:

>NotSerializableException is treated as a transient problem in the TCP based remoting layer. The problem will be logged and message is dropped. Other exceptions will tear down the TCP connection because it can be an indication of corrupt bytes from the underlying transport.

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
```diff
@@ -114,27 +119,31 @@ abstract class CirceAkkaSerializer[Ser <: AnyRef: ClassTag](system: ExtendedActo
...
-        assert(codec._2 != null)
+        codecsMap(parents(clazz.getName))
       } catch {
-        case e @ (_: NoSuchElementException | _: NotSerializableException) =>
-          log.error(
-            e,
-            s"No codec found for [{}] class. Call Register[A] for this class or its supertype and append result to codecs.",
-            clazz.getName)
-        case e: AssertionError =>
+        case e: NoSuchElementException =>
```
PawelLipski, line 129: Nit: maybe a regular `if` instead of try...catch?

PawelLipski (approved):

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
MarconZet: The condition in `if` is verbose:
```
if(!codecsMap.contains(parents.getOrElse(clazz.getName, null)))
```
I think try catch is easier to unserstand

PawelLipski: Ok, let's leave as is

PR #87, opened 2021-09-03
Regex filter annotation for filtering types

### LINKED ISSUE #86: False negative in registration-checker-compiler-plugin
In Hydra (PR \#1167, commit 6b4f0b32a8897e7cd824c57cd986d41d9ab5b80e), I'm getting a 

```
....CirceAkkaSerializer - No codec found for [....] class. Call Register[A] for this class or its supertype and append result to codecs.
java.util.NoSuchElementException: key not found: ...
````

only in runtime (from `org.virtuslab.ash.circe.CirceAkkaSerializer#scanForSerializables`) even though this should've been reported in the compile time. Details on slack coz this refers to proprietary code

### PR DESCRIPTION
**Pull request by MarconZet**
close #86

### DIFF
--- a/registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
+++ b/registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
@@ -2,15 +2,20 @@ package org.virtuslab.ash
 
 import java.io.BufferedWriter
 import java.io.FileWriter
+import java.util.regex.PatternSyntaxException
 
 import scala.annotation.tailrec
 import scala.collection.mutable
+import scala.reflect.ClassTag
+import scala.reflect.classTag
 import scala.tools.nsc.Global
 import scala.tools.nsc.Phase
 import scala.tools.nsc.plugins.PluginComponent
 
 import org.virtuslab.ash.RegistrationCheckerCompilerPlugin.classSweepPhaseName
 import org.virtuslab.ash.RegistrationCheckerCompilerPlugin.serializerCheckPhaseName
+import org.virtuslab.ash.annotation.SerializabilityTrait
+import org.virtuslab.ash.annotation.Serializer
 
 class SerializerCheckCompilerPluginComponent(
     classSweep: ClassSweepCompilerPluginComponent,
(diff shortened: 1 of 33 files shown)

### CODE REVIEW

File: registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala [hunk 1]
```diff
@@ -27,6 +28,7 @@ class SerializerCheckCompilerPluginComponent(
     s"Checks marked serializer for references to classes found in $serializerCheckPhaseName"
 
   private val serializerType = typeOf[Serializer]
+  private val serializerRegexFilter = typeOf[SerializerRegexFilter]
```
PawelLipski, line 31: Hmmm what is the regex specifically matched against? Also, is it "includes" or "fully matches" semantics?

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
```diff
@@ -27,6 +28,7 @@ class SerializerCheckCompilerPluginComponent(
...
   private val serializabilityTraitType = typeOf[SerializabilityTrait]
```
PawelLipski, line 32: While you're here... I'd refine the error message a bit:

`not referenced in marked serializer` => `not registered` (?) `in a class annotated with @org.Serializer. This will lead to a missing codec for Akka serialization in the runtime`

Plus also maybe mention the regex? Since the class might be referenced but not within the desired regex

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
```diff
@@ -27,6 +28,7 @@ class SerializerCheckCompilerPluginComponent(
...
 
   private var typesNotDumped = true
```
PawelLipski, line 34: Also, I'm getting a false positive on the current state of Hydra PR \#1190 (commit `c730950a2`) -- getting `<some-class> not referenced in marked serializer` despite it's located in line 72... oh, maybe the regex should have some extra `.*`s...

PawelLipski: Yeah, works with `.*Register.*`... but not with `Register.*`... does it match the entire source code lines?...

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
MarconZet: Full match

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
MarconZet: It doesn't match source code lines, but `String` representation of `Type`. So `Register` is actually a `org.virtuslab.ash.circe.Register[A]`

PawelLipski: Oh okay... if you could replace `.*Option.*` in the sample test with `scala\.Option\[.*\]` maybe, this will give more clarity (other than the docs ;) )

PawelLipski: Also! maybe instead of having a separate optional annotation... let's make the regex a param of `@Serializer` annotation? Since without specifying the regex, the user is very likely to get a false negative just as we did on Hydra

PawelLipski: Huh am I missing something? I've tried

```scala
@SerializerRegexFilter("org\\.virtuslab\\.ash\\.circe\\.Register\\[.*\\]")
```

but it fails... although `"org\\.virtuslab\\.ash\\.circe\\.Register.*" passes. 

`Register` is an object so I'd expect the type to be sth like `org.virtuslab.ash.circe.Register$` (??)

registration-checker-compiler-plugin/src/test/resources/serializers/CorrectSerializer.scala
```diff
@@ -1,8 +1,9 @@
 package org.random.project
 
-import org.virtuslab.ash.Serializer
+import org.virtuslab.ash.annotation.{Serializer, SerializerRegexFilter}
 
 @Serializer(classOf[SerializableTrait])
+@SerializerRegexFilter(".*Option.*")
 class CorrectSerializer {
```
PawelLipski, line 7: Also! a point valid for Hydra: will the plugin work as expected when the `Register`s are spread across multiple classes? As suggested on a recent Hydra PR (one class with codecs per package aka _module_)

MarconZet: Suppose there are three traits and marker trait:
```scala
@SerializablilityTrait
trait Mark

sealed trait One extends Mark
sealed trait Two extends Mark
sealed trait Three extends Mark
```

This will compile correctly:

```scala
trait Codecs{
  implicit val codec = ???
}

@Serializer
class Serializer extends Codecs {
  lazy val codecs = Seq(Register[One], Register[Two], Register[Three])
}
```

While this will fail:

```scala
trait Registers{
 lazy val register = Seq(Register[One], Register[Two], Register[Three])
}

@Serializer
class Serializer extends Codecs {
  lazy val codecs = this.register
}
```

MarconZet: If there are some in between traits, for example:

```scala
@SerializablilityTrait
trait Mark
trait Trans

sealed trait One extends Trans
sealed trait Two extends Trans
sealed trait Three extends Mark
```

Then this compiles:
```scala
@Serializer
class Serializer{
  lazy val codecs = Seq(Register[Trans], Register[Three])
}
```

This fails:
```scala
@Serializer
class Serializer{
  lazy val codecs = Seq(Register[One], Register[Two], Register[Three])
}
```

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
MarconZet: I used TypeTag for detection, it's more consistent.

MarconZet: > `Register` is an object so I'd expect the type to be sth like `org.virtuslab.ash.circe.Register$` (??)

We detect types, not function calls. What gets detected is `org.virtuslab.ash.circe.Register.Registration[hydra.something.SomethingCommand]`

serializability-checker-library/src/main/scala/org/virtuslab/ash/annotation/Serializer.scala
```diff
@@ -0,0 +1,6 @@
+package org.virtuslab.ash.annotation
+
+import scala.annotation.nowarn
+
+@nowarn("cat=unused")
+class Serializer(clazz: Class[_], regexPattern: String = ".*") extends scala.annotation.StaticAnnotation
```
PawelLipski, line 6: Maybe `typeRegexPattern` for a bit more self-evidency

registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
```diff
@@ -2,15 +2,20 @@ package org.virtuslab.ash
...
 import scala.tools.nsc.Global
 import scala.tools.nsc.Phase
 import scala.tools.nsc.plugins.PluginComponent
 
 import org.virtuslab.ash.RegistrationCheckerCompilerPlugin.classSweepPhaseName
 import org.virtuslab.ash.RegistrationCheckerCompilerPlugin.serializerCheckPhaseName
+import org.virtuslab.ash.annotation.SerializabilityTrait
+import org.virtuslab.ash.annotation.Serializer
 
 class SerializerCheckCompilerPluginComponent(
```
PawelLipski, line 20: Pls publish the entire updated change on Sonatype so that I can test it against Phoenix `develop` (unless you've done so already)

MarconZet: If you want to test it on Hydra, use `sbt publishLocal`. The version will be the same, as if I had published it on Sonatype.

PawelLipski (approved):
Ok, tested the latest commit against Hydra's develop, works fine

PR #89, opened 2021-09-04
Make scalafix actually remove the unused imports

PawelLipski:


build.sbt
```diff
@@ -36,7 +36,7 @@ lazy val commonSettings = Seq(
       "-Ybackend-parallelism",
       "8",
       "-Ywarn-dead-code",
-      "-Ywarn-unused:-imports,_",
```
PawelLipski, line 39: It turns out Scalafix simply relies on warnings about unused imports to perform the cleanup

sbt-dump-event-schema/src/main/scala/org/virtuslab/ash/DumpEventSchema.scala
```diff
@@ -1,11 +1,8 @@
 package org.virtuslab.ash
 
-import scala.language.postfixOps
-
 import better.files._
 import sbt.Keys._
-import sbt.{File => JFile, _}
```
MarconZet, line 7: This was the reason `-imports` flag was specified, didn't know It could be solved like that

MarconZet (approved):
If it compiles, then nothing can be wrong, approving

PR #91, opened 2021-09-07
Changed format of schema dump from json to yaml

MarconZet:


project/Dependencies.scala
```diff
@@ -15,6 +15,7 @@ object Dependencies {
   val betterFiles = "com.github.pathikrit" %% "better-files" % "3.9.1"
   val sprayJson = "io.spray" %% "spray-json" % "1.3.6"
   val scalaCollectionCompat = "org.scala-lang.modules" %% "scala-collection-compat" % "2.5.0"
+  val moultingYaml = "net.jcazevedo" %% "moultingyaml" % "0.4.2"
```
PawelLipski, line 18: Seems a nice lib, shame that it's no longer maintained apparently :/

PawelLipski: But okay to use ofc

sbt-dump-event-schema/src/main/scala/org/virtuslab/ash/DumpEventSchema.scala
```diff
@@ -1,37 +1,40 @@
...
-    } writer.print(JsArray(filteredJsons.toVector).toString(SchemaPrinter))
-  }
-
-  object SchemaPrinter extends SortedPrinter {
-    private val order =
-      Seq("typeSymbol", "name", "annotations", "parents", "fields", "typeName").zip(1 until 10).toMap
-    protected override def organiseMembers(members: Map[String, JsValue]): Seq[(String, JsValue)] = {
-      members.toSeq.sortBy(x => order.getOrElse(x._1, 10))
-    }
+    } writer.print(finalYaml.prettyPrint)
```
PawelLipski, line 37: Actually! whatever the order of attributes gonna be... is it gonna be deterministic b/w invocations?

sbt-dump-event-schema/src/main/scala/org/virtuslab/ash/DumpEventSchema.scala
```diff
@@ -1,37 +1,40 @@
...
+  import DumpEventSchemaYamlProtocol._
   import DumpEventSchemaPlugin.autoImport._
 
   def apply(outputFile: File, inputDirectory: File): Unit = {
-    val jsons = for {
-      file <- inputDirectory.list.toSeq.sorted(File.Order.byName)
-    } yield new String(file.loadBytes).parseJson.asJsObject
+    val typeDefinitions = for {
+      file <- inputDirectory.list.filterNot(_.isDirectory)
+    } yield new String(file.loadBytes).parseJson.convertTo[TypeDefinition]
```
PawelLipski, line 21: Hmmm would it be possible/reasonable to use YAML everywhere, also for the intermediate files?

sbt-dump-event-schema/src/main/scala/org/virtuslab/ash/DumpEventSchema.scala
PawelLipski: Ok, AFAICS is alphabetical

sbt-dump-event-schema/src/test/resources/target.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - name: a
+    typeName: java.lang.String
+  - name: b
+    typeName: scala.Int
+  - name: c
+    typeName: scala.Double
+  typeSymbol: class
+- parents:
+  - org.random.project.Data
+  name: org.random.project.Data.ClassWithAdditionData
```
PawelLipski, line 21: I'd make `name` always first if easily doable for extra readability

sbt-dump-event-schema/src/test/resources/target.yaml
```diff
@@ -0,0 +1,65 @@
...
+  fields:
+  - name: a
+    typeName: scala.collection.immutable.Map[String,Int]
+  - name: b
+    typeName: scala.Option[Int]
+  - name: c
+    typeName: scala.Tuple3[Int,Int,Int]
+  typeSymbol: class
+- name: org.random.project.Data.TraitWithAnnotation
+  parents:
```
PawelLipski, line 53: Whooops looks like the order isn't deterministic, comparing this entry and the one above :/

sbt-dump-event-schema/src/main/scala/org/virtuslab/ash/DumpEventSchema.scala
MarconZet: Not with current yaml libraries. `circle-yaml` depends on `circe` which in turn depends on `cats`. Compiler plugin is an uber jar, meaning we would add `cats` to every compilation run that uses our plugin.

PawelLipski (approved):

PR #104, opened 2021-09-27
Changed base package to org.virtuslab.ash

MarconZet:


sbt-akka-serialization-helper/src/sbt-test/sbt-dumpschema/dependencies/build.sbt
```diff
@@ -0,0 +1,8 @@
+lazy val scala213 = "2.13.6"
```
PawelLipski, line 1: Hmmm are these files added or moved from somewhere else on this PR? 🤔 since I can see no corresponding deleted files

MarconZet: yes, they are for a new test for resolving library dependencies

PawelLipski (approved):

Issue #98, opened 2021-09-16
Remove `circeAkkaSerializable{De,En}coder` from `HydraCirceAkkaSerializer`


PawelLipski:
As discussed on the today catch-up.

Potential effects on ash codebase:
1. `org.virtuslab.ash.circe.CirceAkkaSerializer#generic{De,En}coder` might no longer be needed
2. `@Serializer` might no longer be the best name for this annotation... new better name TBD

PR #109, opened 2021-09-30
Changed Registration from type alias to case class

MarconZet:


PawelLipski (approved):
_Dobra zmiana_ :pis:

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
```diff
@@ -163,7 +165,7 @@ abstract class CirceAkkaSerializer[Ser <: AnyRef: ClassTag](system: ExtendedActo
    */
   protected def genericCodec: Codec[Ser] = Codec.from(genericDecoder, genericEncoder)
 
-  protected def genericEncoder: Encoder[Ser] =
+  private def genericEncoder: Encoder[Ser] =
```
PawelLipski, line 168: This visibility change is unrelated to the tuple=>case class change, right? 🤔

PawelLipski: Isn't it more related to #98 (?)

MarconZet: It's a refactor I wanted to do and decided to do it now, because it is not backward compatible.

PawelLipski: Btw if you could open a PR to Hydra which adapts to all the new changes (probably as v0.3.3 then), after doing #98

PR #112, opened 2021-10-04
Migrated CI to GitHub Actions

MarconZet:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
+name: Scala CI
+
+on:
+  push:
+    branches: [ main ]
+    tags: ["v*"]
+  pull_request:
```
PawelLipski, line 7: Whoops is it a valid yaml? 😯 with no value for the key

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
+name: Scala CI
+
+on:
+  push:
+    branches: [ main ]
+    tags: ["v*"]
```
PawelLipski, line 6: Hmm `v*` vs `*` in release.yml

PawelLipski: Also... any guarantees on relative order of this workflow vs release.yml?

.github/workflows/ci.yml
MarconZet: It most certainly is

.github/workflows/ci.yml
MarconZet: We can use [official semantic regex](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string), but it's quite long.

.github/workflows/ci.yml
```diff
@@ -67,6 +67,28 @@ jobs:
       - name: Running examples
         run: sbt test
         working-directory: examples/event-migration
+
+  publish:
+    needs: [format, test, run-examples]
+    if: github.event_name != 'pull_request'
```
PawelLipski, line 73: Hmmm so `publish` runs on both each `main` build AND git tag build right? does `sbt ci-release` automagically recognize between the two cases (so that it publishes to snapshots or releases)?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Setup Scala
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.11
+
+      - name: Checking formating
+        run: sbt scalafmtCheckAll
+
+      - name: Checking import order
+        run: sbt "scalafixAll --check"
```
PawelLipski, line 26: Actually... AFAIK this runs most (of not all) of `sbt compile` anyway... so I think the benefit of extracting `scalafmtCheckAll` and `scalafixAll --check` into a separate job, executed in parallel with `test`, is probably minuscule (or maybe even the performance impact is negative)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup Scala
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.11
+    
+      - name: Running examples
```
PawelLipski, line 67: Again, worth considering how the overhead of setting up a VM with Scala compares to any gains from parallel execution here

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+        uses: olafurpg/setup-scala@v10
+
+      - name: Publish from CI
+        run: sbt ci-release
+        env:
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
+          PGP_SECRET: ${{ secrets.PGP_SECRET }}
+          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+
```
PawelLipski, line 91: Nit, remove the superfluous empty lines ;)

.github/workflows/ci.yml
MarconZet: Yes

.github/workflows/ci.yml
MarconZet: Yeah, compile depends on `scalafmtCheckAll` and `scalafixAll --check` depends on compile

.github/workflows/ci.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main]
     tags: ["v*"]
-  pull_request
+  pull_request:
```
PawelLipski: Is it `github actions NOT printing an explicit error in case of malformed config file`, vol. 2137? :trollface:

PR #148, opened 2022-02-14
Optimize code

miloradvojnovic:


codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/ClassSweepCompilerPluginComponent.scala
```diff
@@ -15,8 +15,8 @@ class ClassSweepCompilerPluginComponent(options: CodecRegistrationCheckerOptions
   override val runsAfter: List[String] = List("refchecks")
   override def description: String = s"searches for direct descendants of classes annotated with serializability trait"
 
-  val foundTypes: mutable.Buffer[(String, String)] = mutable.ListBuffer()
-  val typesToUpdate: mutable.Buffer[(String, String)] = mutable.ListBuffer()
+  val foundTypes: mutable.Set[(String, String)] = mutable.Set()
+  val typesToUpdate: mutable.Set[(String, String)] = mutable.Set()
```
MarconZet, line 19: `ClassSweep` phase did not have performance problems, so it's natural that optimisations here don't have any effects.

Furthermore this optimisation may be a downgrade. `foundTypes` and `typesToUpdate` are only being appended to, something buffers are made for. But `HashSet` starts with `defaultInitialCapacity` of 16, meaning that multiple resizes are required during runtime.

MarconZet (commented):
Do you have any logs form profiler or just some generic `printf` for measuring which part of the code took the most time? Making changes blindly may lead to counter-intuitive results.

miloradvojnovic:
> Do you have any logs form profiler or just some generic `printf` for measuring which part of the code took the most time? Making changes blindly may lead to counter-intuitive results.

Didn't use any profiler. Do you have any suggestions on which one to use? I changed the code while keeping the implementation valid and nothing improved compilation time.

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/ClassSweepCompilerPluginComponent.scala
miloradvojnovic: I changed that because in the next phase there are `foundTypes.toSet` and `typesToUpdate.toSet`.

PawelLipski:
> Didn't use any profiler. Do you have any suggestions on which one to use?

Pls start with JProfiler! Java VisualVM also seems a natural first choice

miloradvojnovic:
> > Didn't use any profiler. Do you have any suggestions on which one to use?
> 
> Pls start with JProfiler! Java VisualVM also seems a natural first choice

I tried JProfiler and found out that a big amount of time for compilation goes to this part of code:
```
serializerImplDef.collect {
    case x: Tree if x.tpe != null && x.tpe.toString().matches(filterRegex) => x.tpe
  }
```
The reason for this is because we used Tree which is the root class for all Types and it seems to me that's unnecessary so I checked for only several subclasses of Tree. I tried to find the minimal set of classes with having valid tests and also valid compilation for projects that uses ASH. We can discuss it further if there's some class that is also required to be added here.

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
```diff
@@ -108,29 +106,25 @@ class SerializerCheckCompilerPluginComponent(
...
           try {
-            serializerImplDef
-              .collect {
-                case x: Tree if x.tpe != null => x.tpe
-              }
-              .groupBy(_.toString())
-              .filter(_._1.matches(filterRegex))
-              .map(_._2.head)
+            serializerImplDef.collect {
+              case x @ (_: Select | _: GenericApply) if x.tpe != null && x.tpe.toString().matches(filterRegex) => x.tpe
```
PawelLipski, line 110: Okay! so just limiting the types from `Tree` to `Select | GenericApply` shortened the execution time of the compiler phase two-fold right? 🤔

PawelLipski: Actually... could you elaborate why just these two subtypes were enough? is there a case where a false-positive (false-negative?) compiler error happens because some subtype of `Tree` was NOT covered here?

PawelLipski: I.e. I'm asking whether we don't sacrifice **soundness for performance** here!

miloradvojnovic: > Okay! so just limiting the types from `Tree` to `Select | GenericApply` shortened the execution time of the compiler phase two-fold right? 🤔

Yes. Because there are fewer types for checking nullability and regex and also a list of founded types is used in recursion after and since it's smaller it's faster to process it.

miloradvojnovic: > Actually... could you elaborate why just these two subtypes were enough? is there a case where a false-positive (false-negative?) compiler error happens because some subtype of `Tree` was NOT covered here?

I found the minimal set of classes with having successful tests and compilation for the project that uses ASH. Not sure if all possible subtypes of `Tree` are covered.

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/ClassSweepCompilerPluginComponent.scala
MarconZet: please revert it

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
MarconZet: This is not the only time we iterate through the whole tree searching for something, then why is this iteration so expensive?

Converting `Type` to `String`? Or maybe matching regex? Checking for null is rather fast operation.

miloradvojnovic: We have iterations on other places but this is the only place where we get all possible types by matching `Tree` which is the root of all types. I agree they are fast but since they are applied on a very big number of types, compilation time is increased for big projects. Since I couldn't optimize checking nullability and matching regexp I tried to decrease the number of classes on which filtering is applied and that made progress in compilation time.

MarconZet: Yes, but we iterate through all the classes in all the files, not just one, and it takes 1 second, not 40. So something is wrong with the body of collect, not with the amount of classes we process.

Maybe the way we acquire `string` is wrong, or maybe we can prevent unnecessary regex usage. And if all else fails, there is always parallelisation. 

Either way, work is far from done.

PR #154, opened 2022-05-17
Improve compiler error message/153

LukaszKontowski:
1. Added helpful error message for compilation failures caused by using JDK version lower than 11.
2. Updated LICENSE file with 2022 year.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,6 +3,13 @@ import org.scalafmt.sbt.ScalafmtPlugin.autoImport.scalafmtOnCompile
 import sbt.Keys.{semanticdbEnabled, semanticdbVersion, versionScheme}
 import sbt.VirtualAxis.ScalaVersionAxis
 
+initialize ~= { _ =>
+  if (sys.props.getOrElse("java.specification.version", "0").toDouble < 11.0) {
+    throw new IllegalArgumentException(
+      "Project must be built with java version 11 or higher. Current java version will cause failure of compilation.")
+  }
+}
+
 lazy val supportedScalaVersions = List(scalaVersion213, scalaVersion212)
 
 ThisBuild / scalaVersion := supportedScalaVersions.head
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -21,6 +21,8 @@ sonatypeProfileName := "org.virtuslab"
 
 Global / onChangedBuildSource := ReloadOnSourceChanges
 
+(Compile / compile) := (Compile / compile).dependsOn(checkCurrentJavaVersion).value
```
PawelLipski, line 24: Actually.... would it make sense to put the check into `initialize ~= { _ => ... }` block instead? 🤔

build.sbt
```diff
@@ -202,3 +204,13 @@ lazy val dumpPersistenceSchemaCompilerPlugin = (projectMatrix in file("dump-pers
     assembly / assemblyJarName := s"${name.value}_${scalaBinaryVersion.value}-${version.value}.jar", //Warning: this is a default name for packageBin artefact. Without explicit rename of packageBin will result in race condition
     addArtifact(Compile / assembly / artifact, assembly))
   .jvmPlatform(scalaVersions = supportedScalaVersions)
+
+lazy val checkCurrentJavaVersion =
+  taskKey[Unit]("checks whether SBT has been started with proper java version (at least 11)")
+
+checkCurrentJavaVersion := {
+  if (sys.props.getOrElse("java.specification.version", "0").toDouble < 11.0) {
```
PawelLipski, line 212: Does the project compile with no issues on Java 14/17?

LukaszKontowski: Yes, it compiles with no issues with both 14 and 17 - so in my opinion it is ok to say that JDK version should be at least 11 (might be higher)

build.sbt
LukaszKontowski: I replaced this Compile / compile line with initialize block - please take a look

PawelLipski (approved):
:fantastycznie:

PawelLipski:
Btw as you can see, true merges (aka 2-parent merges) are NOT enabled in this repo... see https://slides.com/plipski/git-machete/#/8 and the following slides for rationale

PR #156, opened 2022-05-25
switch from drafts to automated releases

### LINKED ISSUE #101: Create a GitHub release automatically within GH workflows for tags
Surely some reusable GH-provided action can be used for that purpose.

LukaszKontowski:
Closes #101
While still using the same github actions tool (release-drafter) - this change will cause following change:
1) No more Draft Releases will be created after pushing commits to main (i.e. mainly merging PRs)
2) New Releases will be published after pushing git tag to repo ("version" git tags - with names starting with 'v' character) on main branch only.
I tested this flow on my private repo and it works as expected.


.github/workflows/release-drafter.yml
```diff
@@ -2,15 +2,20 @@ name: Release Drafter
```
PawelLipski, line 2: Pls rename the workflow & the file adequately

.github/workflows/release-drafter.yml
```diff
@@ -2,15 +2,20 @@ name: Release Drafter
...
   update_release_draft:
+    if: github.event.base_ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     steps:
-      # Drafts your next Release notes as Pull Requests are merged into "main"
+      # Publishes new Release as new git tag (vX.Y.Z) is pushed to repository
       - uses: release-drafter/release-drafter@v5
         with:
           config-name: release-drafter-config.yml
+          publish: true
```
PawelLipski, line 17: Pls show an example of how such release looks like from your private repo

CONTRIBUTING.md
```diff
@@ -89,11 +89,15 @@ to both `build.sbt` **and** `project/build.sbt` (so that the sbt plugin added in
...
 
-To publish version x.y.z, type in the console
+To publish version x.y.z, type in the console (on main branch)
 ```shell
 git tag vx.y.z
 git push origin vx.y.z
 ```
 The tagged commit is then released to Maven Central.
 
 Note that we are using an early semantic versioning scheme.
```
PawelLipski, line 99: While you're here, pls add a link to explanation what **early** semver means (as my impression is that it's mostly used in Scala world)

.github/workflows/release-drafter.yml
```diff
@@ -2,15 +2,20 @@ name: Release Drafter
 
 on:
   push:
-    branches: [main]
+    tags:
+      - 'v*'
 
 jobs:
   update_release_draft:
+    if: github.event.base_ref == 'refs/heads/main'
```
PawelLipski, line 10: Hmm 🤔 what does `base_ref` mean in context of a workflow executed **for a tag**?

.github/workflows/release-drafter.yml
LukaszKontowski: image

.github/workflows/release-drafter.yml
LukaszKontowski: Hm, I am not quite sure. Tag is related to a commit, not to branch, so it is somehow confusing. I followed an advice from https://github.community/t/run-workflow-on-push-tag-on-specific-branch/17519 which did work for my private repo - it did recognise branch name. Looks like base_ref contained information about branch connected with the commit. But now when I think about it and do some reading - it seems that it does not work always and might not work. Let me remove this line in upcomming commit. We should go without this check.

.github/workflows/release-drafter.yml
LukaszKontowski: done in the newest commit

CONTRIBUTING.md
LukaszKontowski: done in the newest commit

.github/workflows/release-drafter.yml
LukaszKontowski: this `if` condition has been removed in the newest commit

PawelLipski: Huh actually... let's assume for a moment the (likely) scenario when someone pushes a tag for a commit that is NOT a part of `master` history.

How does the behavior differ for such case with and without `if: github.event.base_ref == 'refs/heads/main'`? 

My suspicion is that this `if` might protect us from accidentally releasing a commit that is not a part of master 🤔

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,20 @@
...
+    tags:
+      - 'v*'
+
+jobs:
+  publish-github-release:
+    runs-on: ubuntu-latest
+    steps:
+      # Publishes new Release as new git tag (vX.Y.Z) is pushed to repository
+      - uses: release-drafter/release-drafter@v5
+        with:
```
PawelLipski, line 14: TBD about the `if`, see the old thread

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,20 @@
...
+          config-name: release-drafter-config.yml
```
PawelLipski, line 15: Pls rename this config file accordingly as well

CONTRIBUTING.md
```diff
@@ -89,11 +89,15 @@ to both `build.sbt` **and** `project/build.sbt` (so that the sbt plugin added in
...
-To publish version x.y.z, type in the console
+To publish version x.y.z, type in the console (on main branch)
 ```shell
 git tag vx.y.z
 git push origin vx.y.z
 ```
 The tagged commit is then released to Maven Central.
 
-Note that we are using an early semantic versioning scheme.
+Note that we are using an early semantic versioning scheme (see https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html#early-semver-and-sbt-version-policy for details).
```
PawelLipski, line 99: Verbatim long URLs look bad in the docs, just use
```suggestion
Note that we are using an [early semantic versioning scheme](https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html#early-semver-and-sbt-version-policy).
```

CONTRIBUTING.md
```diff
@@ -89,11 +89,15 @@ to both `build.sbt` **and** `project/build.sbt` (so that the sbt plugin added in
...
+
+### Github Releases
+
+Github Releases are done automatically - with settings defined in the .github/workflows/release-drafter.yml file. Release is published when new "vX.Y.Z" git tag is pushed to the main branch.
```
PawelLipski, line 103: Rename the referenced file (maybe add a relative-path link as well)

.github/workflows/release-drafter.yml
LukaszKontowski: yes, that was the reason I added this `if` - however: it worked well on my private repo but I saw some discussions on the web, that github.event.base_ref might be null when tag is the source of deploying workflow. Let me check that some deeper on my repo - I will try to come back with some solution.

LukaszKontowski: without this `if`- such github release will be created also for 'v*' tags related to commits, that were not merged to main branch

.github/workflows/publish-release.yml
LukaszKontowski: This config-name relates to the tool we are using: https://github.com/marketplace/actions/release-drafter (so we use release-drafter-config.yml to configure release-drafter :p ). But I will change the name.

.github/workflows/release-drafter.yml
PawelLipski: Theoreticaly, if the `if` doesn't work as expected... then you could also add a check that `git rev-parse master` is equal to `git rev-parse HEAD`, directly as a script in the pipeline (and fail if the two hashes are NOT equal)

LukaszKontowski: Ok, I tested it more on my private repo - it showed up that `if: github.event.base_ref == 'refs/heads/main'` works well both for Merge commits and without them (we do not have merge commits in ash now). The event for which there are context properties available - is **push** - and it works well for pushing tags - github is able to recognize base_ref  related to this tag push - which is the branch. So I will just paste this if statement back.

LukaszKontowski: `if` statement added (after tests)

PawelLipski (approved):

Issue #159, opened 2022-06-03
Add support for LZ4 compression


LukaszKontowski:
This issue is a continuation of https://github.com/VirtusLab/akka-serialization-helper/issues/72

Within https://github.com/VirtusLab/akka-serialization-helper/issues/72 compression with Gzip has been added to circe-akka-serializer. However, it would be good to add possibility to use LZ4 algorithm (https://github.com/lz4/lz4-java) instead of Gzip as a second choice (since LZ4 could be faster than GZip). See https://doc.akka.io/docs/akka/current/serialization-jackson.html#compression as a reference / example.

PR #160, opened 2022-06-03
added compression feature for circe-akka-serializer

### LINKED ISSUE #72: Add gzip compression for large payloads
Jackson **Akka serializer** (not Jackson itself!) compresses data, when they are over 32KiB

LukaszKontowski:
Closes #72

Added possibility to use Gzip compression (default setting is "off"). In the future - it would be good to add also LZ4 compression option - this is within  https://github.com/VirtusLab/akka-serialization-helper/issues/159


circe-akka-serializer/src/main/resources/reference.conf
```diff
@@ -1,5 +1,19 @@
 # Default configuration
 
 org.virtuslab.ash {
+
   verbose-debug-logging = off
+
+  # settings for compression of the payload
+  compression {
```
PawelLipski, line 8: I'd extract a sub-section for `circe`, to match package structure in Scala code

circe-akka-serializer/src/main/resources/reference.conf
```diff
@@ -1,5 +1,19 @@
...
+    # Compression algorithm.
+    # - off  : no compression
+    # - gzip : using common java gzip
+    algorithm = off
+
+    # If compression is enabled with the `algorithm` setting,
+    # the payload will be compressed if it's size is bigger than this value.
```
PawelLipski, line 15: `its`

circe-akka-serializer/src/main/resources/reference.conf
```diff
@@ -1,5 +1,19 @@
...
+    compress-larger-than = 0 KiB
```
PawelLipski, line 16: Hmmm regardless of algorithm being `off` by default... shouldn't the default threshold be sth more usable, like 32 KiB? How does it look like in Akka Jackson's reference.conf?

LukaszKontowski: Yeah, you're right - they have 32 KiB as default (see https://doc.akka.io/docs/akka/current/serialization-jackson.html#compression for reference). Let me change that in upcomming PR fixes commit. We'll set it to 32 KiB but leave it "off" for now. Does that make sense?

LukaszKontowski: fixed

circe-akka-serializer/src/main/resources/reference.conf
LukaszKontowski: done

circe-akka-serializer/src/main/resources/reference.conf
PawelLipski: > Does that make sense?

Indeed, that's what I meant

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
```diff
@@ -45,8 +45,18 @@ abstract class CirceAkkaSerializer[Ser <: AnyRef: ClassTag](system: ExtendedActo
     with AkkaCodecs {
 
   private lazy val log = Logging(system, getClass)
-  private lazy val conf = system.settings.config.getConfig("org.virtuslab.ash")
+  private lazy val conf = system.settings.config.getConfig("org.virtuslab.ash.circe")
   private lazy val isDebugEnabled = conf.getBoolean("verbose-debug-logging") && log.isDebugEnabled
```
PawelLipski, line 49: Oh! but `verbose-debug-logging` is NOT related to Circe specifically... it should stay where it used to be!

circe-akka-serializer/src/main/resources/reference.conf
```diff
@@ -1,5 +1,19 @@
 # Default configuration
 
 org.virtuslab.ash {
-  verbose-debug-logging = off
+  circe {
+    verbose-debug-logging = off
```
PawelLipski, line 5: See below

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceAkkaSerializer.scala
LukaszKontowski: ok, I just saw that it has been used only within circe-akka-serializer module - that's why I changed it as well

LukaszKontowski: let me fix that one back to previous state

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceTraitCodec.scala
```diff
@@ -114,6 +105,33 @@ trait CirceTraitCodec[Ser <: AnyRef] extends Codec[Ser] {
...
+  doNeededChecksOnStart()
+
+  private def doNeededChecksOnStart(): Unit = {
+    checkImplementationForInvalidMembersDeclarations()
+    checkSerializableTypesForMissingCodec(packagePrefix)
+    checkCodecsForNull()
+    checkCodecsForDuplication()
+  }
+
+  private def checkImplementationForInvalidMembersDeclarations(): Unit = {
```
PawelLipski, line 124: Nit:
```suggestion
  private def checkImplementationForInvalidMemberDeclarations(): Unit = {
```
https://virtuslab.github.io/tips/#english/grammar-tips-1 point 4

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceTraitCodec.scala
```diff
@@ -85,8 +76,7 @@ trait CirceTraitCodec[Ser <: AnyRef] extends Codec[Ser] {
     .toMap
     .withDefaultValue("")
 
-  def manifest(o: AnyRef): String = parentsUpToRegisteredTypeMap(o.getClass.getName)
-
+  // Decoder apply method - decodes from Json into an object of type Ser
```
PawelLipski, line 79: Nit: I'd even change it into a proper `/** ... */` scaladoc, same below

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/Compression.scala
```diff
@@ -0,0 +1,60 @@
...
+      try readChunk()
+      finally inputStream.close()
+      outputStream.toByteArray
+    } else {
+      bytes
+    }
+
+  private[circe] def isCompressedWithGzip(bytes: Array[Byte]): Boolean =
+    (bytes != null) && (bytes.length >= 2) &&
+    (bytes(0) == GZIPInputStream.GZIP_MAGIC.toByte) &&
```
PawelLipski, line 58: Is it theoretically possible that a payload which is NOT compressed with gzip will still start with these 2 bytes? 🤔

PawelLipski: (and hence incorrectly recognized as compressed)

circe-akka-serializer/src/main/resources/reference.conf
LukaszKontowski: it showed up that this config is relatted only to circe akka serializer and thus it can stay as it is (under circe path)

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/Compression.scala
LukaszKontowski: I am not 100% sure, but this GZIP_MAGIC constant is a standard for all .gz files and also - such check is done in akka-serialization-jackson as well (which is used as a standard for akka cluster). See: https://github.com/akka/akka/blob/main/akka-serialization-jackson/src/main/scala/akka/serialization/jackson/JacksonSerializer.scala#L80

LukaszKontowski: So I think we can assume that this is safe

PawelLipski: Okay... AFAICS `GZIP_MAGIC` is 35615 aka `8b1f` in Hex... both characters being outside of ASCII range.

Could you confirm that **without compression, all characters in the serialized data are from ASCII range**?
If that assumption is correct, I'd even add a test to demonstrate that.

> such check is done in akka-serialization-jackson as well (which is used as a standard for akka cluster)

I wouldn't trust that even if this is safe for `akka-serialization-jackson`, this will still be safe for our serializer... unless the above holds true ☝🏻

LukaszKontowski: Ok, I did some research and debugging with printlns and that's what I've found:
  
1. After my changes there are two ways of converting a serializable object to binary (to Array[Byte]) and it happens inside CirceAkkaSerializer#toBinary method. Interesting part of code in this method is as follows:
```
val bytes = printer.print(encoder.asInstanceOf[Encoder[AnyRef]](o)).getBytes(UTF_8)
val result = Compression.compressIfNeeded(bytes, bufferSize, compressionAlgorithm)
```
  
And compressIfNeeded might do compression or just return raw bytes defined earlier. And we want to be sure, that if serialization is done without compression (raw bytes passed) - then it will not be possible that:
```
isCompressedWithGzip(bytes) == true
```
  
2. So, I started checking what exactly is produced as `bytes` inside CirceAkkaSerializer#toBinary method:
  ```
 val bytes = printer.print(encoder.asInstanceOf[Encoder[AnyRef]](o)).getBytes(UTF_8)
```
`printer.print` "Returns a string representation of a pretty-printed JSON value." (javadoc quoted)
  And afterwards this String is encoded into byte array (`getBytes`) using UTF_8
  
Since we are encoding JSON objects - they will start with '{' char. Thus, the first element of the `bytes` array is always the 123 Byte number - which is the decimal representation of the `{` character (123 decimal is the same as 7b hexadecimal).
  
3. I diged the `GZIP_MAGIC` number - which is a java int with the value of 35615 (8b1f hex). One of the conditions in our test is as below:
```
(bytes(0) == GZIPInputStream.GZIP_MAGIC.toByte)
```
which gets evaluated to:  
```
bytes(0) == 35615.toByte
```
which gets evaluated to:  

```
bytes(0) == 31 // but 31 is of type Byte ...
```
  
And since bytes(0) holds a Byte with value equal to 123 - this will never be true.
  
JFYI - 31 is a decimal representation of the '=' sign
  
So, in my opinion, this will work well and there is no reason to worry. Since the flow goes like this:  
JSON object.   >.   String object.   >.   Array[Byte] object. 
which means that before transforming into byte array, CIRCE takes care of creating Json from our object and Json must not start with '=' character

File: circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceTraitCodec.scala (hunk 7 above)
Comment (LukaszKontowski) [reply circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/CirceTraitCodec.scala]: done for both

File: circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/Compression.scala (hunk 8 above)
Comment (PawelLipski) [reply circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/Compression.scala:58]: > JFYI - 31 is a decimal representation of the '=' sign

Hmmm 31 (aka `1f` in hex) is actually a non-printable character: `UNIT SEPARATOR (US) DOWN ARROW`: https://www.fileformat.info/info/unicode/char/1f/index.htm

But still, that's even less likely to occur in JSON data, let alone as the first character 😅 

Other than that &mdash; a brilliant analysis, thanks! if you could put an **abridged** version of that as a comment in code, so that the future generation don't scratch their heads

Comment (PawelLipski) [reply circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/Compression.scala:58]: > Since we are encoding JSON objects - they will start with '{' char.

For clarity &mdash; we're always encoding JSON **objects**, as in `{...}`, right? can't we ever encode a JSON number/string/array as well?

Comment (LukaszKontowski) [reply circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/Compression.scala:58]: We are encoding objects extending AnyRef type - so it could be either a Json object (starting with '{' character - 123 decimal) or a Json Array (starting with '[' character - 91 decimal) - but this second option would rather be very rare in my opinion. So - for both cases GZIP_MAGIC number will work well.

File: circe-akka-serializer/src/test/resources/application.conf [hunk 9]
```diff
@@ -9,9 +9,23 @@ akka.actor {
...
+  circe {
+    # enables debug logging
+    #verbose-debug-logging = on
+
+    # settings for compression of the payload
+    compression {
+      # Compression algorithm.
+      # - off  : no compression
+      # - gzip : using common java gzip
+      algorithm = off
```
PawelLipski, line 24: One more thing! as we discussed on Slack: when `algorithm = off`, will the deserializer still be able to **read** compressed data? 🤔 this is a hard requirement to cater for rollouts, when for a short time old nodes without compression support might work alongside new nodes with compression support.

PawelLipski: Pls document in reference.conf

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/Compression.scala
LukaszKontowski: Comment, added - please kindly review. :)

circe-akka-serializer/src/test/resources/application.conf
LukaszKontowski: yes, it will be able to read either data compressed with Gzip or uncompressed. I will add description in the reference.conf

circe-akka-serializer/src/main/scala/org/virtuslab/ash/circe/Compression.scala
```diff
@@ -53,6 +53,20 @@ object Compression {
       bytes
     }
 
+  /*
+  Since we are encoding JSON for Ser <: AnyRef types - they will start with '{' char for Json objects or with '[' char for Arrays.
```
PawelLipski, line 57: Slight nuance here... what about `String`s? they extend `AnyRef` as well
Ofc strings in Json start with `"` character, so not a problem here anyway

LukaszKontowski: yes, my mistake - let me add String to this description - " character is represented by 34 decimal so this one is also safe

PawelLipski (approved):

PawelLipski (approved):

PR #166, opened 2022-06-15
Fix false positives for the codec registration checker

### LINKED ISSUE #141: False-positive `No codecs for ... registered` error when switching branches (sth wrong with incremental compilation?)
Waiting for further details from the team. Just the combination of git commit + contents of `target/codec-registration-checker-cache.csv` should be enough.

LukaszKontowski:
Closes #141

- false positives fixed - both for situations, where git branch is changed in the middle of sbt compilations, and for situations where on the same branch new registrations has been added and later deleted
- as usage of https://github.com/pathikrit/better-files has been added to the runtime logic of the codec-registration-checker-compiler-plugin - this dependency has to be provided within a fat jar when ASH gets published - hence changes inside build.sbt
- description with instructions in README added (one new parameter for codec-registration-checker-compiler-plugin)
- added logic - checking source code directory for files containing possible "positives" - gets invoked only if such possible positives were detected - moreover - works quick (just few seconds on Hydra if there are possible missing registrations). So there is no real negative impact on performance.


README.md
```diff
@@ -304,6 +304,13 @@ class ExampleSerializer(actorSystem: ExtendedActorSystem)
 Note that as with Serializability Checker and Dump Persistence Schema,
 this compiler plugin only runs in the sbt modules where `AkkaSerializationHelperPlugin` is explicitly enabled.
 
+To make this plugin work correctly, path to the source code directory where codecs are registered has to be set in sbt project settings - just add settings like below:
+```
+.settings(scalacOptions ++=
+    Seq(s"-P:codec-registration-checker-plugin:--source-code-directory=${baseDirectory.value.file.getAbsolutePath}/path/to/your/code/directory"))
```
PawelLipski, line 310: Huh that'll make the usage pretty complicated 🤔

Could you explain why passing the source code directory is needed to prevent the error?

PawelLipski: Also... even if it stays as is, I'd use a more realistic path as a sample (with `src/main/scala` etc.)

build.sbt
```diff
@@ -56,6 +56,26 @@ lazy val commonSettings = Seq(
       if (sys.env.getOrElse("CI", "false") == "true") "-Xfatal-warnings" else ""),
   libraryDependencies ++= commonDeps)
 
+lazy val assemblySettings = Seq(
```
PawelLipski, line 59: Pls document on why that's needed! (plugin compiler classpath etc.)

PawelLipski: Actually... lemme open a separate task for researching the topic

README.md
LukaszKontowski: > Huh that'll make the usage pretty complicated 🤔
> 
> Could you explain why passing the source code directory explicitly is needed to prevent the error?

Yes - because the only way to check if these missing types are "false-positives" or not - is by checking the source code in the repo that uses ASH plugins. This check does recursive search on all .scala files in the `--source-code-directory` for these possibly missing types. In order not to check whole repository (e.g. several modules in a bigger project) - it would be beneficial to specify the `--source-code-directory` - to check just the directory, where codecs are actually registered in code.

It might be also possible not to specify `path/to/your/code/directory` and to check whole repository by just checking ${baseDirectory.value.file.getAbsolutePath}
It would be one-step-easier for clients but in such situation - all .scala files from whole repo would be checked.

Unfortunately - this check has to be done on source code. Due to how incremental compilation works in sbt - we would not ba able to collect all "really existing" types using scala reflection.

build.sbt
PawelLipski: https://github.com/VirtusLab/akka-serialization-helper/issues/168

README.md
LukaszKontowski: alternatively we can just set `${baseDirectory.value.file.getAbsolutePath}/src/main/scala` as a default ?

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/CodecRegistrationCheckerCompilerPlugin.scala
```diff
@@ -63,10 +75,12 @@ class CodecRegistrationCheckerCompilerPlugin(override val global: Global) extend
         false
     }
   }
-  override val optionsHelp: Option[String] = Some("""
+  override val optionsHelp: Option[String] = Some(s"""
       |. - directory where cache file will be saved, required
-      |--disable - disables the plugin
+      |$disableFlag - disables the plugin
+      |$sourceCodeDirectoryFlag - path of the source code directory, which has to be checked with this plugin
```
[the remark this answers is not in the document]
PawelLipski: Hmmm `src/main/scala/hydra`, or rather `services/src/main/scala/hydra`?

As top-level `src/main/scala/` does not exist:

image

README.md
PawelLipski: (see the other comment about Hydra's case, for which this directory will **not** exist IIUC)

LukaszKontowski: It will work - see my reply on other comment - ${baseDirectory.value.file.getAbsolutePath} evaluates to the module (subproject) directory where plugin is enabled

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/CodecRegistrationCheckerCompilerPlugin.scala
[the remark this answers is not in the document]
PawelLipski: As a small improvement here, I'd extract the path to Scala source code directory programatically from sbt, shouldn't be too difficult (rather than hardcoding `src/main/scala`), let's then think what to do next

README.md
PawelLipski: >  as a default ?

Okay... so will there be cases where ones will **not** want to simply set this flag to `src/main/scala` in the current subproject? 🤔

PawelLipski: Okay! pls proceed

MarconZet: Check out `scalaSource.value`

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
```diff
@@ -176,6 +160,69 @@ class SerializerCheckCompilerPluginComponent(
             None
         }
       }
+
+      private def collectMissingClassNames(fullyQualifiedClassNames: List[String]): List[String] = {
```
MarconZet, line 164: If I may provide an alternative solution to this problem.

For this problem I originally thought about reusing already existing compiler phase.

The non-existing FQCN are resolved by typer compiler phase (you can check this by using `-Ystatistics`)

So if we could create our own typer phase (by `global.analyzer.newTyper`) and send it prepared by plugin AST that contains the missing FQCN the phase would resolve these names for us.

LukaszKontowski: Ok, can you specify why adding our own custom typer phase would be a better solution?

LukaszKontowski: My first thought was - it would be smart but it also adds one more phase to each compilation - whereas my solution adds additional method invocation only in situations, where there were some missing types found. In my opinion we do not need one more phase for this particular reason (these false positives should not be that often? rather a rare case?)

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/CodecRegistrationCheckerCompilerPlugin.scala
```diff
@@ -24,9 +26,19 @@ class CodecRegistrationCheckerCompilerPlugin(override val global: Global) extend
...
 
   override def init(options: List[String], error: String => Unit): Boolean = {
-    if (options.contains("--disable"))
+    if (options.contains(disableFlag))
       return false
 
+    options.find(flag => flag.contains(sourceCodeDirectoryFlag)) match {
+      case Some(directoryFlag) =>
+        pluginOptions.sourceCodeDirectoryToCheck =
+          directoryFlag.substring(24) // as --source-code-directory= is exactly 24 chars long
```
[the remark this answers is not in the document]
LukaszKontowski: done

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/CodecRegistrationCheckerCompilerPlugin.scala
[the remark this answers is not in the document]
LukaszKontowski: done

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
```diff
@@ -176,6 +160,69 @@ class SerializerCheckCompilerPluginComponent(
...
+          val packageName = fqcn.substring(0, indexOfLastDotInFQCN)
+          val typeName = fqcn.substring(indexOfLastDotInFQCN + 1)
+          sourceCodeFilesAsStrings.exists(fileAsString => {
+            fileAsString.startsWith(s"package $packageName") &&
+            (fileAsString.contains(s"class $typeName") || fileAsString.contains(s"trait $typeName") || fileAsString
+              .contains(s"object $typeName"))
+          })
+        }
+
+        for (fqcn <- fullyQualifiedClassNames if typeIsDefinedInScalaFiles(fqcn)) yield fqcn
```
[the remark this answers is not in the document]
LukaszKontowski: done

README.md
LukaszKontowski: I just uploaded working solution - please see https://github.com/VirtusLab/akka-serialization-helper/pull/166/files#diff-d06aaa8abc0aa4f2b1333353fd3962771928ea9140a9ae91b0963b236bf2adceR58

build.sbt
PawelLipski, line 59: Actually! pls add a reference to https://github.com/sbt/sbt/issues/2255 + a short explanation in comment here

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/CacheFileInteractionMode.scala
```diff
@@ -0,0 +1,5 @@
+package org.virtuslab.ash
+
+sealed trait CacheFileInteractionMode
+case class DumpTypesIntoCacheFile() extends CacheFileInteractionMode
```
PawelLipski, line 4: Nit: just case objects?

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
```diff
@@ -176,6 +160,69 @@ class SerializerCheckCompilerPluginComponent(
...
+            fileAsString.startsWith(s"package $packageName") &&
+            (fileAsString.contains(s"class $typeName") || fileAsString.contains(s"trait $typeName") || fileAsString
+              .contains(s"object $typeName"))
+          })
+        }
+
+        fullyQualifiedClassNames.filter(typeIsDefinedInScalaFiles)
+      }
+
+      private def interactWithTheCacheFile(
```
PawelLipski, line 183: Actually, given that this method is only ever called with a specific `mode` known upfront (and not a variable one)... it'll be better to split this method into two independent methods (esp. that the 2 `match` cases are completely separate)

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
```diff
@@ -176,6 +160,69 @@ class SerializerCheckCompilerPluginComponent(
...
+          mode: CacheFileInteractionMode,
+          typeNamesToRemove: List[String] = List.empty): Unit =
+        mode match {
+          case _: DumpTypesIntoCacheFile =>
```
PawelLipski, line 187: In a nutshell, this case is the old piece from `def apply`, while the other case is newly implemented, right?

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/CacheFileInteractionMode.scala
PawelLipski: Or better, just get rid of them completely, as suggested in SerializerCheckCompilerPluginComponent.scala

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
LukaszKontowski: yes

codec-registration-checker-compiler-plugin/src/test/scala/org/virtuslab/ash/CodecRegistrationCheckerCompilerPluginSpec.scala
```diff
@@ -26,7 +28,7 @@ class CodecRegistrationCheckerCompilerPluginSpec extends AnyWordSpecLike with sh
       File.usingTemporaryDirectory() { directory =>
         val out = CodecRegistrationCheckerCompiler.compileCode(
           serializersCode(0) :: dataSourceCode,
-          List(s"${directory.toJava.getAbsolutePath}"))
+          List(s"${directory.toJava.getAbsolutePath}", s"--source-code-directory=$sourceCodeDirectory"))
```
PawelLipski, line 31: Nit: pls add a case where `--source-code-directory` is **not** passed and exception is thrown

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
MarconZet: > method invocation only in situations, where there were some missing types found

This doesn't change regardless of method

MarconZet: > why adding our own custom typer phase would be a better solution?

It doens't use fat jar, better files and also it does not depend on sbt.

build.sbt
```diff
@@ -189,23 +211,6 @@ lazy val dumpPersistenceSchemaCompilerPlugin = (projectMatrix in file("dump-pers
         }
```
PawelLipski, line 211: Actually... would it make sense to mark `scalaPluginDeps21x` as `Provided`? 🤔

PawelLipski: So as to avoid packing at least `scala.*` classes into the fat jar

LukaszKontowski: you mean these ones?
image
I can try to change it if you wish

build.sbt
LukaszKontowski: Sure - comment in file added. Short explanation is in the PR description, but here we go again :)
- as usage of https://github.com/pathikrit/better-files has been added to the runtime logic of the codec-registration-checker-compiler-plugin - this dependency has to be provided within a fat jar when ASH gets published - hence changes inside build.sbt
TL;DR: without publishing fat jar our commercial project (Hydra) did not find classes better-files during compilation - that's why fat-jar was needed

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/CacheFileInteractionMode.scala
LukaszKontowski: File removed - according to other comment (no need to have these types as interactWithTheCacheFile has been replaced with two separate methods)

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
LukaszKontowski: done

codec-registration-checker-compiler-plugin/src/test/scala/org/virtuslab/ash/CodecRegistrationCheckerCompilerPluginSpec.scala
LukaszKontowski: sure, test added

build.sbt
PawelLipski: Yes! I do get it... I meant a comment **in code**, coz locating comments on GitHub via `git blame` isn't particularly convenient :/

PawelLipski: Assume someone would revisit this place in, say, 3 years 😅

LukaszKontowski: Ok, so - is the comment in build.sbt enough - or write some more detailed explanation?

sbt-akka-serialization-helper/src/main/scala/org/virtuslab/ash/AkkaSerializationHelperPlugin.scala
```diff
@@ -51,7 +54,8 @@ object AkkaSerializationHelperPlugin extends AutoPlugin {
         new File(
           (ashDumpPersistenceSchema / ashDumpPersistenceSchemaOutputDirectoryPath).value) / (ashDumpPersistenceSchema / ashDumpPersistenceSchemaOutputFilename).value,
       ashDumpPersistenceSchema / ashDumpPersistenceSchemaOutputFilename := s"${name.value}-dump-persistence-schema-${version.value}.yaml",
-      ashDumpPersistenceSchema / ashDumpPersistenceSchemaOutputDirectoryPath := ashCompilerPluginCacheDirectory.value.getPath) ++
+      ashDumpPersistenceSchema / ashDumpPersistenceSchemaOutputDirectoryPath := ashCompilerPluginCacheDirectory.value.getPath,
+      ashCodecRegistrationCheckerCompilerPlugin / sourceCodeDirectory := s"--source-code-directory=${baseDirectory.value.getAbsolutePath}${File.separator}src${File.separator}main${File.separator}scala") ++
```
PawelLipski, line 58: Okay... so just this one potentially left to sort out... other than that, PR ready to merge I think :)

codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
PawelLipski: (@LukaszKontowski consider opening a separate _self-contained_ issue, i.e. with all known details outlined &mdash; no reference to this thread, for the sake of clarity)

PawelLipski: It's hard to me to assess how significant the suggested improvement is... but I'd intuitively go for 2. as well 😅

LukaszKontowski: Sure, I'll create a new issue for this improvement with detailed description.

build.sbt
LukaszKontowski: done

sbt-akka-serialization-helper/src/main/scala/org/virtuslab/ash/AkkaSerializationHelperPlugin.scala
LukaszKontowski: switched to use `sourceDirectory` - unfortunately - did not find more elegant working solution

sbt-akka-serialization-helper/src/main/scala/org/virtuslab/ash/AkkaSerializationHelperPlugin.scala
```diff
@@ -55,7 +55,7 @@ object AkkaSerializationHelperPlugin extends AutoPlugin {
           (ashDumpPersistenceSchema / ashDumpPersistenceSchemaOutputDirectoryPath).value) / (ashDumpPersistenceSchema / ashDumpPersistenceSchemaOutputFilename).value,
       ashDumpPersistenceSchema / ashDumpPersistenceSchemaOutputFilename := s"${name.value}-dump-persistence-schema-${version.value}.yaml",
       ashDumpPersistenceSchema / ashDumpPersistenceSchemaOutputDirectoryPath := ashCompilerPluginCacheDirectory.value.getPath,
-      ashCodecRegistrationCheckerCompilerPlugin / sourceCodeDirectory := s"--source-code-directory=${baseDirectory.value.getAbsolutePath}${File.separator}src${File.separator}main${File.separator}scala") ++
+      ashCodecRegistrationCheckerCompilerPlugin / sourceCodeDirectory := s"--source-code-directory=${sourceDirectory.value.getAbsolutePath}${File.separator}main${File.separator}scala") ++
```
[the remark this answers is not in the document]
LukaszKontowski: comment added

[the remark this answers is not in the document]
PawelLipski: Actually! for future, unless your reply is non-trivial, you can just `Resolve` the comment without further reply

PR #177, opened 2022-06-28
make release publishing dependent on scala ci

### LINKED ISSUE #172: GitHub release should only be created AFTER the `publish` job is done
In the current setup, release is created in a workflow independent from `Scala CI` workflow :/ We don't want to create a release if e.g. publishing fails due to whatever reason.

LukaszKontowski:
closes #172 
- `publish-github-release` ci job made dependent on results of the whole scala ci workflow
- scala ci workflow pulled out to a separate file and made a reusable workflow


.github/workflows/standard-workflow.yml
```diff
@@ -0,0 +1,161 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out
+        uses: actions/checkout@v2
+
+      - name: Setup Scala
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.11
```
PawelLipski, line 95: Nit: is there some neat way of extracting these repeatable 2 steps (like e.g. in CircleCI: https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/.circleci/config.yml#L2)? if no, let's leave as is

.github/workflows/standard-workflow.yml
```diff
@@ -0,0 +1,161 @@
...
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set up Scala
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.11
+
+      - name: Publish from CI
```
PawelLipski, line 155:
```suggestion
      - name: Publish Maven artifacts
```

.github/workflows/publish-release.yml
```diff
@@ -6,7 +6,11 @@ on:
       - 'v*'
```
PawelLipski, line 6: `name` to be changed to `Publish GitHub Release` for clarity

.github/workflows/publish-release.yml
```diff
@@ -6,7 +6,11 @@ on:
...
 
 jobs:
+  run-scala-ci-release:
+    uses: ./.github/workflows/standard-workflow.yml
```
PawelLipski, line 10: Out of curiosity... isn't it possible to merge this workflow file into what's now `.github/workflows/standard-workflow.yml`, this removing the need for `.github/workflows/standard-workflow.yml`? 🤔

PawelLipski: In other words... is it possible to define `on` on a per-`job` (and not per-workflow) level?

LukaszKontowski: > In other words... is it possible to define `on` on a per-`job` (and not per-workflow) level?

I think that it is not possible - see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on - `on` works on workflow-level.
However, I can try to get all the logic in one `ci.yml` file - and trigger it for:
- pushing to main
- pull request
- pushing tag
And just add one more `if` to the `publish-github-release` job to start only if the source event was a push of a git tag - **if it is possible**

PawelLipski: Not insisting, ignore if not easily doable

LukaszKontowski: But in my opinion the solution from PR - 3 separate files - is more clear and I would prefer it that way (two separate workflows for different triggering events). Let me know what is your opinion on that @PawelLipski

PawelLipski: Okay! let's leave then

.github/workflows/standard-workflow.yml
LukaszKontowski: Looks like it might be a problem: https://github.community/t/support-for-yaml-anchors/16128/88 - so let me leave it as it is for now

Issue #171, opened 2022-06-21
Refactor existing solution for false-positive `No codecs for ... registered` errors


LukaszKontowski:
Currently, in order not to have false-positives (dummy errors) for scenarios like below:
1. On existing project add new top level serializable class and register it in serializer
2. Compile the code (compilation successful)
3. Delete this recently added top level serializable from source code and remove it's registration from serializer (just git restore on files affected by changes from step 1. )
4. Compile the code (compilation fails with the same error)

We have an additional check in the codec-registration-checker-compiler-plugin - see `val actuallyMissingFullyQualifiedClassNames` usage from SerializerCheckCompilerPluginComponent.scala

It works well, but there is a potentially better solution proposed by @MarconZet:

"If I may provide an alternative solution to this problem.

For this problem I originally thought about reusing already existing compiler phase.

The non-existing FQCN are resolved by typer compiler phase (you can check this by using -Ystatistics)

So if we could create our own typer phase (by global.analyzer.newTyper) and send it prepared by plugin AST that contains the missing FQCN the phase would resolve these names for us."

Such solution would allow us to:
- not publish fat-jar for this plugin (as the only reason we do this is a problem with better-files dependency that has to be bundled)
- not depend on sbt
- not depend on client's source code in such a direct way (currently we check source code in a somehow barbaric way - see `private def collectMissingClassNames` inside SerializerCheckCompilerPluginComponent.scala)

The idea of this issue is to check if suggested alternate solution is possible and (if it is) implement this as a refactor from existing solution.

PR #180, opened 2022-07-01
fixed  logic of `collectMissingClassNames`

### LINKED ISSUE #179: Fix `collectMissingClassNames` logic - codec-registration-checker-plugin
It shows up that current logic from `collectMissingClassNames` method does not work well in particular:
```
sourceCodeFilesAsStrings.exists(fileAsString => {
  fileAsString.startsWith(s"package $packageName") &&
  (fileAsString.contains(s"class $typeName") || fileAsString.contains(s"trait $typeName") || fileAsString
     .contains(s"object $typeName"))
})
```

It does not work well for scenarios where serializable trait is nested in an object, example:
https://github.com/VirtusLab/akka-on-kubernetes/blob/exercise-5-solution/server/src/main/scala/com/virtuslab/akkaonk8s/tictactoe/GameActor.scala

LukaszKontowski:
closes #179 
Removed invalid assumption about package names. Assumption was wrong as serializable traits can be nested.


codec-registration-checker-compiler-plugin/src/main/scala/org/virtuslab/ash/SerializerCheckCompilerPluginComponent.scala
```diff
@@ -204,12 +204,11 @@ class SerializerCheckCompilerPluginComponent(
 
         def typeIsDefinedInScalaFiles(fqcn: String): Boolean = {
           val indexOfLastDotInFQCN = fqcn.lastIndexOf('.')
-          val packageName = fqcn.substring(0, indexOfLastDotInFQCN)
           val typeName = fqcn.substring(indexOfLastDotInFQCN + 1)
           sourceCodeFilesAsStrings.exists(fileAsString => {
-            fileAsString.startsWith(s"package $packageName") &&
-            (fileAsString.contains(s"class $typeName") || fileAsString.contains(s"trait $typeName") || fileAsString
-              .contains(s"object $typeName"))
+            fileAsString.contains(s"class $typeName") ||
```
PawelLipski, line 209: Huh so the condition is now relaxed... it's getting out of hand TBH 😅

Pls add a test and merge as is... but generally #171 will be very welcome if doable 🤔

LukaszKontowski: Ok, I will do #171 soon then. Regarding regression testing - please take a look at two existing tests that we have. The first one ensures that plugin detects missing registrations. The second one ensures that we do not treat "false positives" as errors. Do we need another test case? If yes - what should be tested?

https://github.com/VirtusLab/akka-serialization-helper/blob/main/codec-registration-checker-compiler-plugin/src/test/scala/org/virtuslab/ash/CodecRegistrationCheckerCompilerPluginSpec.scala#L55-L63

and

https://github.com/VirtusLab/akka-serialization-helper/blob/main/codec-registration-checker-compiler-plugin/src/test/scala/org/virtuslab/ash/CodecRegistrationCheckerCompilerPluginSpec.scala#L55-L63

PawelLipski: I'd specifically like a test that would've capture this bug! I.e. a one that would fail before this fix but won't fail anymore after the fix

PawelLipski: That'll be useful esp. given that the logic is likely to be rewritten as part of #171

LukaszKontowski: Test case added - "type that is missing in serializer definition is wrapped in an object" + small refactor in the test class to improve readability

PawelLipski (approved):

PR #183, opened 2022-07-07
Refactor false positives fix

### LINKED ISSUE #171: Refactor existing solution for false-positive `No codecs for ... registered` errors

LukaszKontowski:
Closes #171 

- switched from direct checks on files in user's source code to usage of `global.findMemberFromRoot` method
- stopped relying on better-files in the published code
- stopped publishing fat jars for codec registration checker compiler plugin

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,7 +299,7 @@ class ExampleSerializer(actorSystem: ExtendedActorSystem)
     // ...
     override lazy val codecs = Seq(Register[CommandOne]) // WHOOPS someone forgot to register CommandTwo...
     // ... but Codec Registration Checker will throw a compilation error here:
-    // `No codec for `CommandOne` is registered in class annotated with @org.virtuslab.ash.annotation.Serializer`
+    // `No codec for `CommandOne` is registered in a class annotated with @org.virtuslab.ash.annotation.Serializer`
 }
 ```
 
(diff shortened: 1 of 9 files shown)


PawelLipski (approved):
Looks good to me! 🎉

Issue #199, opened 2022-07-12
Provide config for scala-steward


PawelLipski:
As per https://github.com/scala-steward-org/scala-steward/blob/1d020673731dbe2c73f5be233a8599ac79f70adf/docs/repo-specific-configuration.md

Some things that I can see:
* update dependencies in `examples/` which is a separate sbt project (`buildRoots`)
* run `scalafixAll` as `postUpdateHook` after scalafix sbt plugin or scalafix rules (just `organize-imports` currently) are updated
* `updatePullRequests`: maybe `always`?

fthomas:
> run `scalafixAll` as `postUpdateHook` after scalafix sbt plugin or scalafix rules (just `organize-imports` currently) are updated

Running `scalafixAll` in case of organize-imports updates sounds interesting. I'm not familiar with organize-imports but assume that `scalafixAll` will organize the imports in projects that depend on that rule. IMO that sounds like a post-update hook that could be added the [default Scala Steward post-update hooks](https://github.com/scala-steward-org/scala-steward/blob/b48a850c1b76a09cdbbc6ceaf88d2d47f60498c3/modules/core/src/main/resources/default.scala-steward.conf#L1-L15) so that is enabled in all projects that use Scala Steward.

PawelLipski:
> that could be added the [default Scala Steward post-update hooks](https://github.com/scala-steward-org/scala-steward/blob/b48a850c1b76a09cdbbc6ceaf88d2d47f60498c3/modules/core/src/main/resources/default.scala-steward.conf#L1-L15)

Yeah indeed, scalafix sounds like a much more common tool than e.g. `sbt-java-formatter` that's included there 😅

aluscent:
@PawelLipski I'm on it. 🤟🏻
I was a lil bit busy these days. Let's rock.

PawelLipski:
Gr8!

aluscent:
@PawelLipski This is the new [.scala-steward.conf](https://github.com/aluscent/akka-serialization-helper/blob/main/.scala-steward.conf) file.
I've scheduled it to create PRs every 5 days. There's also an `@asap` tag to create PRs immediately as there was an update available, but I think it can have so much overhead for the bot.
I've also left the post-update hooks empty.
I don't exactly know your requirements, so if you need I can brief you about all of the configurations.
Also, please tell me if you need to have any dependency ignored, or if you want a specific dependency to be updated according to some version pattern.

[...]

PR #209, opened 2022-07-27
Adding .scala-steward.conf file

aluscent:
According to #199


.scala-steward.conf
```diff
@@ -0,0 +1,103 @@
...
+# It is possible to have multiple scala projects in a single repository. In that case the folders containing the projects (build.sbt folders)
+# are specified using the buildRoots property. Note that the paths used there are relative and if the repo directory itself also contains a build.sbt the dot can be used to specify it.
+# Default: ["."]
+buildRoots = [ "." ]
+
+# Define commands that are executed after an update via a hook.
+# A groupId and/or artifactId can be defined to only execute after certain dependencies are updated. If neither is defined, the hook runs for every update.
+postUpdateHooks = [
+  {
+    command = ["sbt", "scalafixAll"],
```
PawelLipski, line 87: Hmmm this is probably too broad... will run `sbt scalafixAll` after an update of **any** dependency, right? Note that `scalafixAll` basically needs to compile the entire project.

See https://github.com/scala-steward-org/scala-steward/pull/2671 ... I think we can rely on defaults here; our `postUpdateHooks` can probably be empty

.scala-steward.conf
```diff
@@ -0,0 +1,103 @@
...
+# Default: @asap
+#
+#pullRequests.frequency = "0 0 ? * 3" # every thursday on midnight
+pullRequests.frequency = "5 days"
+
+# Only these dependencies which match the given patterns are updated.
+#
+# Each pattern must have `groupId`, and may have `artifactId` and `version`.
+# Defaults to empty `[]` which mean Scala Steward will update all dependencies.
+updates.allow  = []
```
PawelLipski, line 34: This entire file is based on some blueprint from scala-steward right? I think that re-stating the comments and defaults here is redundant, one can just check them in the upstream .conf file... I'd remove the comments, and only include the settings that are actually changed from the default!

.scala-steward.conf
```diff
@@ -0,0 +1,103 @@
...
+# If "always", Scala Steward will always update the PR it created as long as
+# you don't change it yourself.
+# If "never", Scala Steward will never update the PR
+# Default: "on-conflicts"
+updatePullRequests = "always"
+
+# If set, Scala Steward will use this message template for the commit messages and PR titles.
+# Supported variables: ${artifactName}, ${currentVersion}, ${nextVersion} and ${default}
+# Default: "${default}" which is equivalent to "Update ${artifactName} to ${nextVersion}"
+commits.message = "Scala-Steward: Update ${artifactName} from ${currentVersion} to ${nextVersion}"
```
PawelLipski, line 70: Actually... do you know how it'll deal with multiple deps getting updated in a single PR? no need to check it now tho if you don't, we'll just see that with the first such update

.scala-steward.conf
```diff
@@ -0,0 +1,103 @@
...
+
+# If true and when upgrading version in .scalafmt.conf, Scala Steward will perform scalafmt
+# and add a separate commit when format changed. So you don't need reformat manually and can merge PR.
+# If false, Scala Steward will not perform scalafmt, so your CI may abort when reformat needed.
+# Default: true
+scalafmt.runAfterUpgrading = false
```
PawelLipski, line 76: Huh why override to `false`? 🤔 `true` seems the reasonable default here

.scala-steward.conf
```diff
@@ -0,0 +1,103 @@
...
+
+# It is possible to have multiple scala projects in a single repository. In that case the folders containing the projects (build.sbt folders)
+# are specified using the buildRoots property. Note that the paths used there are relative and if the repo directory itself also contains a build.sbt the dot can be used to specify it.
+# Default: ["."]
+buildRoots = [ "." ]
```
PawelLipski, line 81: `examples/` are another build root! see #199

.scala-steward.conf
```diff
@@ -0,0 +1,103 @@
...
+#     minutes, hour of day, day of month, month, and day of week.
+#
+#     See https://www.alonsodomin.me/cron4s/userguide/index.html#parsing for
+#     more information about the CRON expressions that are supported.
+#
+#     Note that the date parts of the CRON expression are matched exactly
+#     while the time parts are only used to abide to the frequency of
+#     the given expression.
+#
+# Default: @asap
```
PawelLipski, line 25: Hmmm it seems that this setting does **not** affect the runs of Scala Steward itself... only just how often the PRs are created. Creating PRs is just a minor part of the workload for Scala Steward for this repo... the majority is probably checking for dependency updates themselves.

I'd leave the default `@asap` here... I think non-`@asap` settings only make sense in repos where there would be a constant flood of new PRs otherwise. This is clearly **not** the case for this project: we haven't got a single update from Scala Steward in ~2 weeks now, since none of our dependencies has released a new version recently.

aluscent:
Thank you @PawelLipski for your feedbacks, I will fix them ASAP!

PawelLipski (approved):
Looks good! this will probably result in a few new PRs once Scala Steward runs again... b/c of `examples/` being included now

.scala-steward.conf
```diff
@@ -0,0 +1,5 @@
+updatePullRequests = "always"
+
+commits.message = "Scala-Steward: Update ${artifactName} from ${currentVersion} to ${nextVersion}"
+
+buildRoots = [ ".", "examples/" ]
```
PawelLipski, line 5:
```suggestion
buildRoots = [ ".", "examples/" ]

```

PR #218, opened 2022-08-02
Handle possible OverlappingFileLockExceptions for cache file

### LINKED ISSUE #217: Laguna compilation fails with ASH versions greater than 0.4.3
`sbt compile` of Laguna repo fails for all ASH versions since 0.4.4 - example logs:
```
[error] Exception thrown, message: null
[error] one error found
[error] (edgeAsset / Compile / compileIncremental) Compilation failed
[error] Total time: 82 s (01:22), completed Aug 1, 2022, 5:22:48 PM
```

We should debug the cause of failure.
- If the cause is a bug in Akka Serialization Helper - this bug should be fixed in this issue
- If the cause is not a bug in Akka Serialization Helper - help Laguna team recognize what is wrong and fix that + add some advice for users in either README or GUIDE if needed

LukaszKontowski:
closes #217 
Description added in a comment to the while loop.

### DIFF
--- a/docs/GUIDE.md
+++ b/docs/GUIDE.md
@@ -119,7 +119,9 @@ This option disables detection of messages/events/state based on return type of 
 `Compile / scalacOptions += "-P:serializability-checker-plugin:--disable-detection-higher-order-function"`<br><br>
 
 ### Codec Registration Checker Compiler Plugin
-Before using this compiler plugin, make sure that you are using both [annotations](#annotations) properly. If so &mdash; the plugin can be used right away. This plugin checks whether classes marked with serializability trait are being referenced in a marked serializer, which ensures that codecs will be registered in runtime.<br><br>
+Before using this compiler plugin, make sure that you are using both [annotations](#annotations) properly. If so &mdash; the plugin can be used right away. This plugin checks whether classes marked with serializability trait are being referenced in a marked serializer, which ensures that codecs will be registered in runtime.
+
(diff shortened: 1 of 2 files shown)


docs/GUIDE.md
```diff
@@ -119,7 +119,9 @@ This option disables detection of messages/events/state based on return type of
...
 
 ### Codec Registration Checker Compiler Plugin
-Before using this compiler plugin, make sure that you are using both [annotations](#annotations) properly. If so &mdash; the plugin can be used right away. This plugin checks whether classes marked with serializability trait are being referenced in a marked serializer, which ensures that codecs will be registered in runtime.<br><br>
+Before using this compiler plugin, make sure that you are using both [annotations](#annotations) properly. If so &mdash; the plugin can be used right away. This plugin checks whether classes marked with serializability trait are being referenced in a marked serializer, which ensures that codecs will be registered in runtime.
+
+**Note** - Codec Registration Checker Compiler Plugin is useful only in projects (modules) where the [@Serializer](#Serializer) annotation is used. Therefore, if you are using Akka Serialization Helper in multiple modules but in fact use `@Serializer` annotation in only one module, you might disable this plugin in all other modules except the one where `@Serializer` annotation is used.<br><br>
```
PawelLipski, line 124: Actually... isn't there an sbt key to disable this compiler plugin? that's easier than passing `--disable` flag explicitly 🤔

PawelLipski: > Looks like these sbt flags are more elegant solution - it would be nice to have such flags for all possible scalac options for our compiler plugins. 

Okay, pls open but as a `nice to have` feature. As for the flags for which there are already sbt keys... pls update the GUIDE accordingly now (I mean, can be on a separate PR)

PawelLipski (approved):

docs/GUIDE.md
LukaszKontowski: sure, separate PR for this change will come soon today

PR #237, opened 2022-08-24
enable Dependabot security reports

### LINKED ISSUE #236: Enable security reports from `Dependabot`
Enabling `Dependabot` security reports would increase security.

See https://www.scala-lang.org/blog/2022/07/18/secure-your-dependencies-on-github.html

LukaszKontowski:
Closes #236


.github/workflows/dependency-graph.yml
```diff
@@ -0,0 +1,12 @@
...
+  push:
+    branches:
+      - main
+jobs:
+  dependency-graph:
+    name: Update Dependency Graph
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: scalacenter/sbt-dependency-submission@v2
```
PawelLipski, line 12: Okay... so what are going to be the developer-visible effects of this action? an email notification if some of the dependencies is flagged as vulnerable?

.github/workflows/dependency-graph.yml
```diff
@@ -0,0 +1,12 @@
+name: Update Dependency Graph
+on:
+  push:
+    branches:
+      - main
+jobs:
+  dependency-graph:
+    name: Update Dependency Graph
```
PawelLipski, line 8: Okay so dependency graph does NOT update in GitHub automatically? or just there isn't a built-in support in dependency graph for Scala?

LukaszKontowski: > We released [sbt-dependency-submission](https://github.com/marketplace/actions/sbt-dependency-submission), a GitHub action that submits the dependencies of sbt builds to the [GitHub Dependency submission API](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/using-the-dependency-submission-api) (currently in beta). This action, once installed in a GitHub workflow, allows you to view the [Dependency graph](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph) in GitHub and to receive regular security reports from [Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts).

from the https://www.scala-lang.org/blog/2022/07/18/secure-your-dependencies-on-github.html article.

Looks like Dependabot by default supports static configuration files, such as POM.xml. And `scalacenter/sbt-dependency-submission@v2` makes it possible to use `sbt` with `Dependabot` together

.github/workflows/dependency-graph.yml
LukaszKontowski: https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts#access-to--dependabot-alerts

> By default, we notify people with admin permissions in the affected repositories about new Dependabot alerts

We can configure / disable notifications. 

Also:
> Dependabot alerts are displayed on the Security tab for the repository and in the repository's dependency graph. The alert includes a link to the affected file in the project, and information about a fixed version.

LukaszKontowski: looks like e-mail by default - to admins + we have it visible on the Security tab

.github/workflows/dependency-graph.yml
LukaszKontowski: For now, under https://github.com/VirtusLab/akka-serialization-helper/network/dependencies - we have only .yml files from github actions

PawelLipski (approved):

PR #267, opened 2022-09-26
Mention ScalaObjectDeserializerModule from jackson serializer in Readme

### LINKED ISSUE #266: In README document jackson ScalaObjectDeserializerModule
Most of what you say in the readme about Jackson is correct. One item that yo could note is this:

If you add ScalaObjectDeserializerModule in Jackson config, you will not get new instances of Scala objects being created - this module is not included in DefaultScalaModule for backward compatibility reasons. It will be part of DefaultScalaModule in jackson-module-scala 3.0.0 release.

LukaszKontowski:
Closes #266


LukaszKontowski:
Diff is not easy to read at all. The real only change is adding the:
> (without adding ScalaObjectDeserializerModule usage)
part in the appendix

PawelLipski (approved):

